PR #10, opened 2026-01-29
feat(flows): add validate command for flow file validation

Ben8t:
- Introduced a new command `validate` to validate flow definitions from YAML files or directories.
- Implemented error handling for missing arguments and nonexistent files.
- Added tests for the new command to ensure proper functionality and error reporting.
- Removed error logging from the main function for cleaner output.
- Added example in README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,12 @@ kestra flows deploy ./flows/ --fail-fast
 
 # Combine flags
 kestra flows deploy ./flows/ --namespace prod --override --fail-fast
+
+# Validate a single flow
+kestra flows validate path/to/flow.yaml
+
+# Validate all flows in a directory (recursive)
+kestra flows validate ./flows/
 ```
 
 ### Executions
--- a/main.go
+++ b/main.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"fmt"
 	"os"
 
 	"github.com/kestra-io/kestra-cli/src/cli"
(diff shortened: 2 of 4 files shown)


Ben8t:
@AcevedoR LMK what do you think but the `formatValidateResults` is fairly long mainly because:
The SDK response is minimal and indexed, which forces the CLI to do extra work:
- It returns a flat list of ValidateConstraintViolation entries keyed by an index instead of a per‑flow structure, so we must map indices back to files.
- It can return multiple entries per flow (and multiline constraint strings), so we normalize and dedupe for readable output.
- It doesn’t differentiate “fatal vs non‑fatal” semantics; the CLI must interpret constraints as failure and warnings/infos as non‑fatal.


I already raised an issue on validation endpoint being a bit "dumb" from a end user perspective  https://github.com/kestra-io/kestra/issues/13367 (this one has been disqualified by @kestra-io/backend because from a server perspective it all make sense)

AcevedoR:
Putting code in SDK vs CLI is a tradeoff, I would argue it's better in the CLI because the project is cleaner and way more easily maintenable.

I'm starting to wonder if we should not create an other validate endpoint, that is designed for API users and outputs a cleaner report, BUT, still uses the same validation a the current endpoint (designed for our UI) WDYT @Ben8t ? should not be hard to impl

Ben8t:
Yes... Florian/Loic made it clear the current API endpoint for validation is good and should not change (at least in its response mode). Not sure having a new endpoint will help, as it will make things even less clear.
For the CLI let's use the code I made (manage error/constraint/warning message per flow); will see if other users raise the points around the API/SDK itself so we validate their is a real l challenge here

AcevedoR (approved):

PR #14, opened 2026-02-02
feat(iam): add IAM commands to manager users, roles, groups and corresponding bindings

### LINKED ISSUE #12: Support IAM management via CLI
Since the Kubernetes Operator will be deprecated, we need to provide programatic support to end users to create and manage IAM resources in Kestra. 

This includes, but is not limited to operations on the following:
- Users
- Groups
- Service Accounts
- Roles
- Bindings
- API Tokens
- Invitations

Ben8t:
closes https://github.com/kestra-io/kestra-cli/issues/12


AcevedoR (commented):
when you can, please take the time to add an e2e test like:
- random user creation with dev role + list a permitted namespace = assert 200
- random user creation with no role + list a permitted namespace = assert 403
etc

Malaydewangan09:
Issue is completed and closed!

PR #27, opened 2026-02-16
feat: add install inscript

AcevedoR:


AcevedoR:
need it to be public @Ben8t

PR #32, opened 2026-02-20
fix: validate nsfile namespace existence via search

Ben8t:


dhileephan-ravindran-1995:
Waiting for this Fix for a implementation.

fdelbrayelle:
Closing as superseded — the underlying issue (namespace existence check misreporting on OSS) is already fixed on `main` at `src/cli/nsfiles.go` `namespaceExists()`, via a different approach: treating OSS's 2xx-decode-error response as "exists" rather than switching to `SearchNamespaces`. Thanks for the fix, but it's no longer needed against current main.

Issue #52, opened 2026-05-29
feat(iam): manage users via CLI


jymaire:
Part of #12 — IAM management via CLI. One PR for this resource.

Add a `kestractl users` command group backed by the Go SDK's `UsersAPI` (instance-level, not tenant-scoped).

### Scope (SDK-supported operations)
- `users list` — `ListUsers`
- `users get <id>` — `User`
- `users create` — `CreateUser`
- `users update <id>` — `UpdateUser` / `PatchUser`
- `users delete <id>` — `DeleteUser`
- `users set-groups <id>` — `UpdateUserGroups`
- `users set-password <id>` — `PatchUserPassword`
- API tokens (folded in here): `users tokens list|create|delete` — `ListApiTokensForUser` / `CreateApiTokensForUser` / `DeleteApiTokenForUser`

### Acceptance criteria
- [ ] `newUsersCommand()` + pure `runUsers...()` functions in `src/cli/users.go`
- [ ] Registered in `root.go`
- [ ] Table + JSON output; `validateOutputFormat()` called early
- [ ] SDK errors wrapped with `formatSDKError`
- [ ] Credentials masked under `--verbose`
- [ ] Unit tests in `src/cli/users_test.go` targeting `runUsers...()`
- [ ] Follows CLAUDE.md / CONTRIBUTING.md command pattern

> Note: requires Kestra EE.

PR #56, opened 2026-05-29
feat: add IAM user commands

jymaire:
## What

Adds a `kestractl users` command group for IAM user management (Kestra EE), backed by the SDK's `UsersAPI`. Implements #52.

### Commands
- `users list` — `--query`, `--page`, `--size`, `--sort`; table + JSON
- `users get <id>`
- `users create` — `--email` (required), `--first-name/--last-name/--user-password/--group/--tenant-grant/--superadmin/--restricted`
- `users update <id>` — partial-safe (see below)
- `users delete <id>` — confirmation prompt, skippable with `-y/--yes`
- `users set-groups <id>` — `--group` (repeatable); no flags clears membership
- `users set-password <id>` — `--user-password`
- `users tokens list|create|delete` — `create` reveals the full token once

## Notes / design decisions
- **`--user-password`, not `--password`** — avoids shadowing the global basic-auth `--password` persistent flag (a collision would otherwise send the new user's password as the auth credential).
- **Partial-safe update** — `superAdmin`/`restricted` are only sent when the operator actually passes the flag (tracked via `Flags().Changed`). The SDK serializes any non-nil `*bool`, so always-sending would silently demote a super-admin on an unrelated update.
- **Clearing groups works** — `set-groups` with no `--group` sends `"groupIds": []` (a non-nil empty slice) rather than omitting the field.
- **JSON output contract honored** — mutating commands (delete/set-groups/set-password/token delete) emit a JSON status object under `--output json`; the confirmation prompt goes to stderr so it never pollutes stdout.
- **Empty lists render `[]`**, not `null`.
- Credentials masked under `--verbose` (root.go now masks any flag whose name contains `token`/`password`).

## Testing
- `go build`, `go vet` clean
- `go test ./src/...` — 169 pass (incl. regression tests for the collision fix, partial-update, group clearing, JSON output, empty-list rendering)
- Manually QA'd against a local Kestra EE instance

## Out of scope (follow-ups)
- Advanced endpoints (impersonate, super-admin patch, refresh-token, auth-method deletion)
- `set-groups` tenant scoping is implicit (uses active tenant); `tokens list` total reflects page length

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+# Compiled binary (built via `go build -o kestractl`)
+/kestractl
+
 # Python-generated files
 __pycache__/
 *.py[oc]
@@ -21,5 +24,3 @@ wheels/
 .idea
 
 .planning/.*
-
-.claude/settings.local.json
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Kestra CLI
 
-A Go-based command-line interface for managing Kestra flows, executions, and namespaces.
+A Go-based command-line interface for managing Kestra flows, executions, namespaces, and IAM users.
 
 ## Installation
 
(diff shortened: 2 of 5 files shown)


jymaire:
## QA Report — live run against Kestra EE v1.3.20

Ran the `users` command surface against a local Kestra EE **v1.3.20** instance (basic auth as super-admin, tenant `main`).

### Summary: 18/18 reachable cases pass; 1 bug found & fixed; 1 case not testable on this instance

| Case | Result |
|------|--------|
| create / get / list / `list -o json` | ✅ PASS |
| empty list → `[]` (not `null`) | ✅ PASS |
| set-password | ✅ PASS |
| tokens create (full token shown once) / list / `delete -o json` | ✅ PASS |
| create requires `--email` | ✅ PASS |
| verbose masks `--user-password` | ✅ PASS |
| delete: `n` cancels / `--yes` deletes / `-o json` clean stdout | ✅ PASS |
| **partial `update` preserves attributes** | ✅ PASS (after fix) |
| `set-groups` | ⚠️ Not testable on this instance (see below) |

### 🐞 Bug found: partial `update` silently demoted users (now fixed)

`users update <id> --first-name X` flipped `superAdmin: true → false` and wiped `lastName`/`groups`/`tenants`. Captured on the wire:

```
PUT /api/v1/users/<id>
{"email":"…","firstName":"Renamed"}          ← superAdmin correctly omitted
→ {"…","superAdmin":false,"groups":[],"tenants":[]}
```

**Root cause:** the endpoint is a true **full-replace PUT** — any field absent from the body is reset server-side. Omitting an unchanged field clears it rather than preserving it.

**Fix:** `runUsersUpdate` now **fetch-merges** — `GET`s the current user, overlays only the explicitly-passed flags, then `PUT`s the complete object. `--email` is no longer required on `update`. Added regression unit tests.

Verified live after the fix:
```
before:  superAdmin=true  lastName=Keep
after `update --first-name Renamed`:  superAdmin=true  firstName=Renamed  lastName=Keep
```
✅ super-admin preserved · ✅ lastName preserved · ✅ firstName applied · ✅ explicit `--superadmin=false` still demotes.

### ⚠️ `set-groups` — not testable here (env limitation + known follow-up #7)

`set-groups` calls the tenant-scoped `PUT /api/v1/main/users/{id}/groups`. On this instance the tenant surface isn't reachable for the super-admin: a fresh instance-level user 404s ("User does not exist…"), `--tenant-grant main` fails ("Tenant does not exist for id 'main'"), and `/api/v1/tenants` + `/api/v1/main/groups` return 401. The CLI builds the correct request (`{"groupIds":[]}` to the right URL), but the behavior can't be exercised on this box.

This reinforces the earlier review note that `set-groups` is hard-coupled to the active tenant — proposed follow-up: an explicit `--tenant` override and a clearer error when the user isn't a tenant member.

> Note: the `update` fetch-merge fix + new tests are pending commit to this branch.

fdelbrayelle (approved):
Next, logs?

jymaire:
> Next, logs?

There are other sub part for IAM  (see sub issues of  #12 ), but yes, I will do it soon afte

PR #77, opened 2026-06-21
ci: Slack alerts for CI failures and release notification

Malaydewangan09:
- Part of kestra-io/kestra-ee#8369

PR #80, opened 2026-06-22
[backport releases/v1.3.x] New CLI commands from SDK coverage expansion

jymaire:
## Backport to `releases/v1.3.x`

Backports the command/feature work merged to `main` since the 1.3 release branch diverged (2026-06-17), so the new CLI commands ship on the Kestra 1.3 line.

### Contents
- **117 feature/fix/docs commits** — the new command groups and command additions (`feat(...)`, `fix(...)`, plus README/AGENTS docs). All cherry-picked with `-x`, zero conflicts.
- **1 CI commit** — `ci: Slack alerts for CI failures and release notification (#77)`, applied cleanly (additive Slack steps only; the `v*` tag trigger and 1.3 pinning are untouched).

### Not included
- `ci: trigger release on v-prefixed tags (#74)` — already present on the release branch via #75 (cherry-pick was empty).

### Verification
- `go build` ✅
- `go test ./src/...` ✅


jymaire:
## Kestra 1.3 compatibility test

Spun up **`kestra-ee:v1.3-no-plugins` (resolved to 1.3.23 EE)** locally and drove **all 215 new leaf commands** this backport adds (exact command-tree diff vs `releases/v1.3.x`), capturing each call's real HTTP status. On this server an unmapped route returns `401`+HAL while a valid route with a missing resource returns `404`/`400`/`422` with a Kestra error message — that distinguishes a genuine endpoint incompatibility from a missing-resource/validation response.

### Result: 214 / 215 compatible

One command targets an endpoint that does not exist in 1.3 and has been **removed** in 76edb60 (latest commit on this branch):

- **`dashboards defaults`** → `GET /api/v1/main/dashboards/settings/default-dashboards` returns `401`/unmapped on 1.3 (endpoint added after 1.3). Removed the command, its run/`orNone` helpers, its unit test, and the README example. All other `dashboards *` commands work.

### Robustness notes (compatible, kept as-is)

These endpoints exist and return `200` with their required parameter, but the CLI doesn't enforce it, so calling them bare makes the 1.3 server throw a `500` (NPE) instead of a clean validation error. Not endpoint incompatibilities, but worth tightening in a follow-up:

- **`server generate`** — needs `--from` (else `from is null` → 500)
- **`test-suites last-result`** — needs `--id` (else `testSuiteIds() is null` → 500)

Build + unit tests green after the removal.

fdelbrayelle (approved):

Issue #83, opened 2026-07-01
kestractl v1.16.0: Cannot deserialize flow labels — json: cannot unmarshal array into Go struct field


timurkhafizov:
## Summary

All `kestractl flows` commands (`deploy`, `get`, `list`) fail with a deserialization error when the target namespace contains flows that use array-format labels. The Kestra API returns labels as `[{"key": "...", "value": "..."}]` (JSON array of objects), but the Go client struct expects a map type.

## Error message

```
API error: json: cannot unmarshal array into Go struct field _FlowWithSource.labels of type kestra_api_client._MapObjectObject
```

## Environment

- **kestractl version:** v1.16.0
- **Kestra server:** v1.3.24
- **OS:** macOS (darwin/arm64)

## Reproduction

Any flow with labels in the standard YAML array format triggers this:

```yaml
labels:
  - key: type
    value: data_extraction
  - key: version
    value: v2
```

Commands that fail:

```bash
kestractl flows deploy my-flow.yml --override
# FAILED: json: cannot unmarshal array into Go struct field _FlowWithSource.labels...

kestractl flows get my.namespace my-flow
# Error: json: cannot unmarshal array into Go struct field _FlowWithSource.labels...

kestractl flows list my.namespace
# Error: json: cannot unmarshal array into Go struct field _Flow.labels...
```

## Impact

- **`flows deploy --override`** never reaches the PUT - update is never issued.
- **`flows get`** and **`flows list`** are completely broken for any namespace that has a flow with labels.
- Only namespaces with zero labelled flows work normally.

## Workaround

Direct API calls bypass the issue:

```bash
curl -s -u USER:PASS -X PUT \
  "http://KESTRA_HOST/api/v1/TENANT/flows/NAMESPACE/FLOW_ID" \
  -H "Content-Type: application/x-yaml" \
  --data-binary @flow.yml
```

## Affected struct fields

- `_FlowWithSource.labels` (triggered by `deploy`, `get`)
- `_Flow.labels` (triggered by `list`)

jymaire:
Hello @timurkhafizov  a fix has been push with today [release](https://github.com/kestra-io/kestractl/releases/tag/v1.16.1) , feel free to re open the issue if the bug is still there

Pradumnasaraf:
It's working now. QA done ✅. CC @timurkhafizov

timurkhafizov:
[...]

PR #86, opened 2026-07-02
test(flows): add regression coverage for FORM-typed inputs

brian-mulier-p:
## Summary
- Regression test for the originally reported bug: `flows list`/`ls` erroring with `FORM is not a valid Type` when a flow has a FORM input, because the pinned go-sdk's `Type` enum was closed and predated `FORM`.
- `list`/`ls`/`get`/`deploy` already tolerated this decode failure via PR #83's raw-body recovery (`tryParseFlowFromError`/`tryParseFlowListFromError`/`tryParseFlowSearchFromError`), independent of the bug it was originally written for.
- `flows search`, `flows revisions`, and `flows bulk-update` were **not** covered by that recovery and needed a real dependency fix: pinned `go-sdk` to a pseudo-version off client-sdk's `main` branch (`go get github.com/kestra-io/client-sdk/go-sdk@main`), which carries the FORM decode fix and tracks Kestra develop — the same axis this branch's e2e tests run against. No client-sdk release was needed.
- Deliberately did **not** pin to `go-sdk v1.3.0`: that tag is cut from client-sdk's `releases/v1.3.x` branch, which tracks Kestra 1.3.x's narrower REST surface (missing endpoints like `triggers create-backfill`/`flows expressions`/`dashboards defaults`, differently-shaped bulk-operation responses, and a different apps-search query format) — the wrong axis for a branch tested against `develop`.
- Added a query-string assertion to `TestRunAppsCatalog` covering the `filters[field][op]=value` shape apps-search sends.
- Added a `go.mod` comment warning against `go get -u`: pseudo-versions sort below tagged releases, so an unguarded upgrade would silently swap this back to `v1.3.0`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,12 @@ module github.com/kestra-io/kestractl
 go 1.25
 
 require (
-	github.com/kestra-io/client-sdk/go-sdk v1.1.0
+	// Pseudo-version of client-sdk/go-sdk@main (tracks Kestra develop). Do not
+	// `go get -u` this: v1.3.0 is a real tag from releases/v1.3.x (Kestra 1.3.x's
+	// API) and sorts higher than this pseudo-version, so an unguarded upgrade
+	// would silently swap back to the wrong API surface. Re-resolve deliberately
+	// with `go get github.com/kestra-io/client-sdk/go-sdk@main`.
+	github.com/kestra-io/client-sdk/go-sdk v1.1.1-0.20260702143038-8c3851bea2e1
 	github.com/posthog/posthog-go v1.10.0
 	github.com/spf13/cobra v1.8.0
 	github.com/spf13/pflag v1.0.10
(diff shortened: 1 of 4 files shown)


jymaire (approved):

go.mod
```diff
@@ -3,7 +3,12 @@ module github.com/kestra-io/kestractl
 go 1.25
 
 require (
-	github.com/kestra-io/client-sdk/go-sdk v1.1.0
+	// Pseudo-version of client-sdk/go-sdk@main (tracks Kestra develop). Do not
+	// `go get -u` this: v1.3.0 is a real tag from releases/v1.3.x (Kestra 1.3.x's
+	// API) and sorts higher than this pseudo-version, so an unguarded upgrade
+	// would silently swap back to the wrong API surface. Re-resolve deliberately
+	// with `go get github.com/kestra-io/client-sdk/go-sdk@main`.
+	github.com/kestra-io/client-sdk/go-sdk v1.1.1-0.20260702143038-8c3851bea2e1
```
jymaire, line 11: This is not what we defined as process so please remove this change

jymaire (changes requested):

go.mod
jymaire: ok, need to add doc to comply with it. Goal is to run kestractl from main branch and not from released version

jymaire (approved):

PR #90, opened 2026-07-07
feat(namespaces): support managing namespace variables

### LINKED ISSUE #89: Support managing namespace variables
## Problem

kestractl has no way to create or update variables on a namespace. The only related command is `namespaces inherited-variables`, which is **read-only** and only shows variables inherited from parent namespaces. Setting variables directly on a namespace currently requires Terraform or the UI/API directly.

## Use case

Some teams manage namespace-level configuration outside of Terraform, from within their own deployment pipelines. Namespace variables are a better fit than the KV store for long-lived configuration values that should live alongside the namespace definition, so being able to set them programmatically via the CLI matters for that workflow.

## Proposed solution

The Kestra Go SDK's `Namespace` model already exposes a `Variables map[string]interface{}` field (with `GetVariables`/`SetVariables`), and it round-trips through the existing `NamespacesAPI.Namespace()` (GET) and `NamespacesAPI.UpdateNamespace()` (PUT) calls — no new SDK or API surface is required.

Suggested CLI additions:
- `kestractl namespaces update <namespace_id> --variable key=value` (repeatable) to set/merge variables
- Surface `variables` in `kestractl namespaces get <namespace_id>` output (table + JSON)
- Consider a `--variables-file` flag for bulk definition (YAML/JSON), consistent with similar flags elsewhere in the CLI

## References

- `src/cli/namespaces.go` — `runNamespacesUpdate`, `runNamespacesInheritedVariables`
- SDK: `go-sdk/kestra_api_client/model_namespace.go` — `Variables` field

fdelbrayelle:
## Summary

- Add `--variable key=value` (repeatable) and `--variables-file` (YAML/JSON) flags to `namespaces create` and `namespaces update` to set namespace variables
- Surface `variables` in `kestractl namespaces get` output (table + JSON)
- Reuses the SDK's existing `Namespace.Variables` field via the current `Namespace()` (GET) / `CreateNamespace()` / `UpdateNamespace()` calls — no new API/SDK surface needed
- `--variable` entries take precedence over `--variables-file` entries on key conflicts
- Setting variables replaces the full variable set on that namespace, matching how `create`/`update` already behave for other fields

Closes #89

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l` clean
- [x] `go test ./src/...` (764 passed)
- [x] Added unit tests: `parseVariableFlags` (pairs, file, override precedence, invalid input), `runNamespacesCreate`/`runNamespacesUpdate` request-body assertions, `runNamespacesGet` output assertions

### DIFF
--- a/README.md
+++ b/README.md
@@ -411,6 +411,11 @@ kestractl namespaces create my.namespace
 kestractl namespaces update my.namespace --description "Production namespace"
 kestractl namespaces delete my.namespace
 
+# Set namespace variables (repeatable --variable, or a YAML/JSON file); replaces
+# the full variable set on that namespace
+kestractl namespaces create my.namespace --variable env=prod --variable region=eu
+kestractl namespaces update my.namespace --variables-file variables.yml
+
 # View inherited secrets and variables
 kestractl namespaces inherited-secrets   my.namespace
 kestractl namespaces inherited-variables my.namespace
--- a/src/cli/namespaces.go
+++ b/src/cli/namespaces.go
@@ -325,6 +325,10 @@ func newNamespacesUpdateCommand() *cobra.Command {
 		Short: "Update a namespace.",
 		Long: `Update a namespace.
 
+Fields not passed on the command line keep their current value: the
+existing namespace is fetched first, and only the flags provided here are
+applied on top before saving.
+
 --variable and --variables-file set the full list of namespace variables,
 replacing any variables previously set on the namespace. Combine them to
 layer inline overrides on top of a file: --variable entries win on key
@@ -347,7 +351,7 @@ conflicts.`,
 			if err != nil {
 				return err
 			}
-			return runNamespacesUpdate(client, args[0], description, variables, renderer)
+			return runNamespacesUpdate(client, args[0], description, cmd.Flags().Changed("description"), variables, renderer)
 		},
 	}
 
@@ -357,9 +361,18 @@ conflicts.`,
 	return cmd
 }
 
-func runNamespacesUpdate(client *Client, id, description string, variables map[string]interface{}, renderer *Renderer) error {
-	ns := kestra.NewNamespace(id, false)
-	if description != "" {
+func runNamespacesUpdate(client *Client, id, description string, descriptionSet bool, variables map[string]interface{}, renderer *Renderer) error {
+	// UpdateNamespace is a full-replace PUT: start from the current namespace
+	// so fields not passed on this invocation aren't wiped.
+	ns, _, err := client.API.NamespacesAPI.Namespace(client.Ctx, id, client.Tenant).Execute()
+	if err != nil {
+		return formatSDKError(err)
+	}
+	if ns == nil {
+		ns = kestra.NewNamespace(id, false)
+	}
+
+	if descriptionSet {
 		ns.SetDescription(description)
 	}
 	if variables != nil {
--- a/src/cli/namespaces.go
+++ b/src/cli/namespaces.go
@@ -3,10 +3,12 @@ package cli
 import (
 	"fmt"
 	"os"
+	"strings"
 	"text/tabwriter"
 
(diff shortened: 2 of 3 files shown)


jymaire:
### ⚠️ CONFIRMED bug: a partial `update` silently wipes namespace variables

**Update:** I've now reproduced this end-to-end against a live Kestra EE v1.3. The namespace update is a full-replace PUT, so any field not passed on the current invocation is cleared.

**Root cause (CLI side):** `runNamespacesUpdate` builds a fresh `Namespace` from only the flags passed on this invocation. A description-only update therefore sends a body with no `variables` key. Captured request body:

```
update my.ns --description hi
→ {"deleted":false,"description":"hi","id":"my.ns"}   # no "variables" key
```

**End-to-end reproduction (live Kestra EE):**

```
1. namespaces create qa.finding1 --variable env=prod --variable region=eu
2. namespaces get    qa.finding1  →  "variables": {"env":"prod","region":"eu"}   ✅ present
3. namespaces update qa.finding1 --description "touched by a partial update"     (no variable flags)
4. namespaces get    qa.finding1  →  "variables": null                            ❌ WIPED
```

The symmetric case holds too: a variables-only update (`--variable ...` with no `--description`) drops a previously-set description.

**Impact:** any user who manages variables with this command and later runs `update` for a different field loses their variables with no warning. This is an easy footgun in scripted/CI usage.

**Suggested fixes (pick one):**
- GET the existing namespace, merge the flag-provided fields onto it, then PUT (preserves untouched fields — matches PATCH-like expectations).
- Or, if full-replace is intended, document it loudly in `--help`/README and warn when `update` is run without the full field set.

Deterministic regression test for the CLI-side mechanism (payload omits unspecified fields):

```go
func TestFinding1_DescriptionOnlyUpdate_OmitsVariables(t *testing.T) { /* asserts no "variables" key in body */ }
func TestFinding1_VariablesOnlyUpdate_OmitsDescription(t *testing.T)  { /* asserts no "description" key in body */ }
```

jymaire (changes requested):
see comment about loss of data

fdelbrayelle:
## QA Report

Ran manual QA against a live Kestra EE instance (Docker, `develop` tag, via internal `plugin-devtools`), exercising the new `namespaces create/update` variable flags end-to-end.

**Setup note (infra, unrelated to this PR):** a fresh EE Docker instance needs `kestra.security.super-admin.tenant-admin-access` set, otherwise the superadmin JWT has no role bindings and every API call 403s (no tenant/ADMIN-role auto-bootstrap happens). Not a kestractl issue, just documenting for future EE QA setups.

### Results

- `--variable k=v` (repeatable) on `create` → persisted correctly, shown in table (`VARIABLES:` block) and JSON output.
- `--variables-file` on `update` → full replace, both YAML and JSON file formats parse correctly.
- `--variable` + `--variables-file` combined → `--variable` wins on key conflict, matches documented precedence.
- Omitting `--description` on `update` clears it — confirmed this is pre-existing PUT-replace behavior (not introduced by this PR).
- Invalid `--variable` (missing `=`) and a missing `--variables-file` path both fail with clear error messages and non-zero exit code.
- No-variables case: `create`/`get` still work; table omits the `VARIABLES:` block, JSON shows `"variables": null` — no regression.

No defects found. LGTM from a QA standpoint.

fdelbrayelle:
Fixed in c829791.

`runNamespacesUpdate` now fetches the current namespace first (`GET`), applies only the flags passed on this invocation on top of it, then sends the full merged object on the `PUT`. A description-only update no longer wipes variables, and vice versa.

Added regression tests: `TestRunNamespacesUpdate_DescriptionOnlyPreservesVariables`, `TestRunNamespacesUpdate_VariablesOnlyPreservesDescription`, `TestRunNamespacesUpdate_GetError`.

jymaire (approved):

PR #95, opened 2026-07-30
feat(plugins): download a single plugin by Maven coordinates  #65

### LINKED ISSUE #65: feat(plugins): download a single plugin by Maven coordinates
## Summary

Add a kestractl command to download **one specific plugin JAR** by its Maven
coordinates (`groupId:artifactId:version`), instead of the whole compatibility
set. This lets users grab or pin a single plugin into their `plugins/` directory
without pulling every plugin for a Kestra version.

## Current behaviour

`kestractl plugins download <kestra-version>` (in `src/cli/plugins.go`) fetches the
**full** plugin list for a Kestra version from the compatibility API
(`api.kestra.io/v1/plugins/artifacts/core-compatibility`) and downloads every JAR
from Maven Central. There is no way to download a single plugin.

## Motivation

- Targeted setups (custom images, slim deployments) need only a handful of plugins,
  not the full set — downloading everything is wasteful and slow.
- Pinning/upgrading one plugin to a specific version is a common operation today done
  by hand (constructing the Maven URL and `curl`-ing the JAR).
- The per-artifact download + verification logic already exists (`downloadJAR`,
  `mavenJARURL`, `pluginFileName`, `fetchMavenSHA1`/`localFileSHA1`/`isValidZIP`), so
  this is mostly coordinate parsing + a thin command wrapper.

## Proposed command

```
kestractl plugins get <groupId:artifactId:version>
```

Example:
```
kestractl plugins get io.kestra.plugin:plugin-aws:0.20.0
```

> **Naming is open for discussion.** `plugins get` keeps the existing bulk
> `plugins download <version>` unambiguous. Alternatives: `plugins download-one`,
> `plugins add`, or overloading `plugins download` to detect a coordinate-shaped arg
> (contains `:`) vs a version arg — though overloading is more error-prone. Maintainer
> call.

### Flags

| Flag | Reuse from `download`? | Purpose |
|---|---|---|
| `--plugins-dir` (default `./plugins`) | yes | Destination directory |
| `--force-redownload` | yes | Re-download even if present and checksum-valid |

`--concurrency` and `--edition` are not relevant for a single artifact and should be omitted.

## Implementation notes

- Parse `groupId:artifactId:version` into the existing `pluginArtifact` struct
  (validate exactly 3 colon-separated, non-empty parts; clear error otherwise).
- Call the existing `downloadJAR(p, pluginsDir, forceRedownload)` — it already handles
  the Maven URL, SHA-1 verification, ZIP magic-byte fallback, and skip-if-up-to-date.
- Reuse `resolveVersion` so `develop`/`latest` aliases work consistently with the bulk command.
- Print a single-line result (`done (X.X MB)` / `already up to date` / `FAILED`) mirroring
  the bulk command's output style; non-zero exit on failure.
- Mark the command `Annotations: {AnnotationOffline: "true"}` like the sibling commands.

## Acceptance criteria

- [ ] `kestractl plugins get <group:artifact:version>` downloads the single JAR into `--plugins-dir`
- [ ] Invalid coordinates produce a clear, actionable error (not a panic / raw HTTP error)
- [ ] Existing, checksum-valid JAR is skipped unless `--force-redownload` is set
- [ ] `develop`/`latest` version aliases resolve via `resolveVersion`
- [ ] Non-zero exit code when the download fails (e.g. artifact not found on Maven Central → HTTP 404)
- [ ] Unit tests covering: coordinate parsing (valid + malformed), happy-path download, skip-when-present, and failure — using an `httptest` server via the existing `pluginsMavenBase`/`pluginsAPIBase` override vars
- [ ] `go build ./...`, `go vet ./...`, `gofmt`/`goimports` clean

## Test plan

- Point `pluginsMavenBase` at an `httptest` server (pattern already used by existing tests),
  assert the JAR is written with the correct `pluginFileName` and that re-running skips it.
- Manual: `kestractl plugins get io.kestra.plugin:plugin-aws:0.20.0 --plugins-dir /tmp/p` and
  confirm the JAR lands and validates.

akashvijay98:
closes https://github.com/kestra-io/kestractl/issues/65.

## Summary

- Adds `plugins get <groupId:artifactId:version>` command to download a single plugin JAR by its Maven coordinates instead of the whole compatibility set
- Parses and validates exactly 3 colon-separated parts and reuses the existing download logic to fetch the individual JAR from Maven Central
- Supports skipping already up-to-date JARs unless `--force-redownload` is set, and allows custom destination via `--plugins-dir` (defaults to `./plugins`)
- Prints a single-line result on success (`done (size)`, `already up to date`) and returns a descriptive error with non-zero exit on failure; runs fully offline

## Test plan

- [ ] `TestParsePluginCoordinate` — verifies coordinate parsing (valid and malformed) and rejects version aliases like `latest`/`develop`
- [ ] `TestRunPluginsGet_HappyPath` — asserts the JAR is successfully fetched via `httptest` server and written with the correct filename
- [ ] `TestRunPluginsGet_SkipsExistingValidJAR` and `TestRunPluginsGet_ForceRedownloadsExistingJAR` — asserts existing JARs are skipped unless the force flag is provided
- [ ] `TestRunPluginsGet_DownloadFailure` and `TestRunPluginsGet_InvalidCoordinate` — covers HTTP 404 failures and actionable invalid coordinate errors
- [ ] `TestPluginsGetCommand_Flags` and `TestPluginsGetCommand_RequiresArgs` — asserts flags exist and Cobra argument validation rejects empty calls
- [ ] Full suite: `go test ./...` — all tests green


MilosPaunovic:
Just a small note for the future, if you could use **GitHub Keywords** when opening PRs to connect it to the proper issue (to automate the entire process a bit more), it would be super helpful to us. You can read a bit more about it [here](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests#linking-a-pull-request-to-an-issue) (I've added closing keyword for this particular issue).

src/cli/plugins.go
```diff
@@ -415,6 +441,52 @@ func runPluginsInstall(out io.Writer, kestraVersion string, pluginsDir string, c
...
+	p, err := parsePluginGetCoordinate(coordinate)
+	if err != nil {
+		return err
+	}
+
+	if err := os.MkdirAll(pluginsDir, 0o755); err != nil {
+		return fmt.Errorf("cannot create plugins directory %q: %w", pluginsDir, err)
+	}
+
+	ctx := context.Background()
```
jymaire: please add a timeout for the download. [example](https://github.com/akashvijay98/kestractl/blob/ff64dca954b55e0f27d2d65c8ec5721b39d8fd8c/src/cli/plugins.go#L332 )

src/cli/plugins.go
```diff
@@ -415,6 +441,52 @@ func runPluginsInstall(out io.Writer, kestraVersion string, pluginsDir string, c
 	return nil
 }
 
+func parsePluginGetCoordinate(coordinate string) (pluginArtifact, error) {
```
jymaire: parsePluginGetCoordinate reimplements parsePluginCoordinates's logic with different validation. Extract one shared  parsePluginCoordinate and have both use it

src/cli/plugins.go
```diff
@@ -415,6 +441,52 @@ func runPluginsInstall(out io.Writer, kestraVersion string, pluginsDir string, c
...
+	if errors.Is(err, errRateLimited) {
+		fmt.Fprintf(out, "%s ... FAILED (rate limited by repository)\n", label)
+		return errRateLimited
+	} else if err != nil {
+		fmt.Fprintf(out, "%s ... FAILED (%v)\n", label, err)
+		return fmt.Errorf("download failed: %w", err)
+	} else if skipped {
+		fmt.Fprintf(out, "%s ... already up to date\n", label)
+	} else {
+		fmt.Fprintf(out, "%s ... done (%.1f MB)\n", label, float64(n)/(1024*1024))
```
jymaire: Download errors are printed as a FAILED line and nil is returned. Scripts and CI can't detect failure. Should return fmt.Errorf(...)

src/cli/plugins.go
```diff
@@ -415,6 +441,52 @@ func runPluginsInstall(out io.Writer, kestraVersion string, pluginsDir string, c
...
+	}
+
+	ctx := context.Background()
+	logf := func(format string, args ...any) {
+		fmt.Fprintf(out, format, args...)
+	}
+
+	label := fmt.Sprintf("%s:%s:%s", p.GroupID, p.ArtifactID, p.Version)
+
+	n, skipped, err := downloadJAR(ctx, logf, p, pluginsDir, forceRedownload, pluginsMavenBase, "", "", nil)
```
jymaire: private/mirrored repos are unreachable with the current version of the PR as there is maven username/password

src/cli/plugins_test.go
```diff
@@ -976,3 +976,189 @@ func TestPluginsListCommand_Flags(t *testing.T) {
...
+		{
+			name:       "valid standard coordinate",
+			coordinate: "io.kestra.plugin:plugin-aws:0.20.0",
+			want:       pluginArtifact{GroupID: "io.kestra.plugin", ArtifactID: "plugin-aws", Version: "0.20.0"},
+			wantErr:    false,
+		},
+		{
+			name:       "valid coordinate with latest alias",
+			coordinate: "io.kestra.plugin:plugin-aws:latest",
+			want:       pluginArtifact{GroupID: "io.kestra.plugin", ArtifactID: "plugin-aws", Version: "999.999.999"},
```
jymaire: 999.999.999` is never released as a version so it should not be send to Maven as a version. Code to adapt in `plugins.go`

jymaire (changes requested):
Thanks for the PR! I added a couple of things to do before to merge it

src/cli/plugins_test.go
akashvijay98: Hi @jymaire, thanks for catching this!
In the current code, the resolveVersion() function maps those aliases to 999.999.999.
To resolve this, I am planning to intercept the latest/develop aliases before the download step. I would then query the artifact's maven-metadata.xml to dynamically resolve the actual latest release version and inject that into the Maven JAR URL instead.

Does this approach sound good or is there a different way you'd prefer I handle version resolution here?

src/cli/plugins.go
akashvijay98: Hi @jymaire Thanks for pointing this out! 
I moved the maven plugin parsing logic into a single parsePluginCoordinate method and both parsePluginCoordinates and runPluginsGet methods now use.

src/cli/plugins.go
akashvijay98: Done. The mavenUsername and mavenPassword credentials are now passed into the downloadJAR.

src/cli/plugins.go
akashvijay98: Fixed. I've updated the implementation to use context.WithTimeout and added a --global-timeout flag.

src/cli/plugins.go
akashvijay98: Hi @jymaire, I wanted to clarify on this one, The code is returning fmt.Errorf when a download fails. Can you tell me if there is a specific edge case which I missed where it is returning nil

MilosPaunovic:
If you've addressed any of the requested changes from the PR review, @akashvijay98, make sure to click the `Resolve conversation` button for that particular change to signal to the reviewer that you've completed what was requested.

src/cli/plugins.go
jymaire: sorry mistake on my side for this one, mis read

src/cli/plugins_test.go
jymaire: Hello @akashvijay98 , latest release might be tough to find because latest release is not always compatible with the Kestra instance and might cause some tricky bugs. The easiest thing to do in this PR would be to reject "latest" with a clear error, asking the user to provide an exact version. (and for EE plugins, artefacts are not on Maven Central so I'm not sure we can use `maven-metadata.xml`)

And then if another PR/release we try to handle the "latest" cases but we break things, it would be easy to rollback than just ship all at once.

Thanks again for your contribution :)

akashvijay98: Thanks for the clarification @jymaire! I've added a validateCoordinateVersion function to explicitly reject plugin coordinate aliases  like 'latest' or 'develop'. I've wired this up inside parsePluginCoordinate. This way, it immediately catches invalid coordinates for the get command without affecting other parts of the codebase, which can continue to safely rely on resolveVersion

src/cli/plugins.go
```diff
@@ -482,18 +562,47 @@ func fetchPluginList(kestraVersion string, license string) ([]pluginArtifact, er
...
+		return fmt.Errorf("version %q is not supported — please specify an exact version (e.g. 1.2.3)", version)
+	default:
+		return nil
+	}
+}
+
+// parsePluginCoordinate parses a single Maven-style plugin coordinate string
+// in the format "groupId:artifactId:version" into a pluginArtifact struct.
+// It returns an error if the coordinate is malformed or missing required parts.
+func parsePluginCoordinate(coord string) (pluginArtifact, error) {
```
Malaydewangan09, line 592: Coordinate parts aren't sanitized, so a value like `io.kestra:../../../tmp/evil:1.0` resolves outside `--plugins-dir` via `filepath.Join`, and the same parts flow unescaped into the Maven URL. Real coordinates never contain path separators, so rejecting any part containing `/`, `\`, or `..` here is a cheap guard. Low severity since the coordinate is a local arg, but worth it for scripts feeding untrusted input.

src/cli/plugins.go
```diff
@@ -415,6 +450,51 @@ func runPluginsInstall(out io.Writer, kestraVersion string, pluginsDir string, c
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), globalTimeout)
+	defer cancel()
+	logf := func(format string, args ...any) {
+		fmt.Fprintf(out, format, args...)
+	}
+
+	label := fmt.Sprintf("%s:%s:%s", p.GroupID, p.ArtifactID, p.Version)
+
+	n, skipped, err := downloadJAR(ctx, logf, p, pluginsDir, forceRedownload, effectiveMavenBase, mavenUsername, mavenPassword, parsedHeaders)
```
Malaydewangan09, line 497: `downloadJAR` does `os.Create` (which truncates the real destination) then `io.Copy`. If the copy fails partway (network drop, or the timeout firing mid-transfer) a truncated `.jar` is left behind. On the next run `--force-redownload` defaults to false, so the existence check passes and it reports "already up to date", and Kestra then loads a broken plugin that survives re-runs. Safer to download to a `.part` temp file in the same dir and rename only after a clean copy and close, removing it on error. Pre-existing shared code, but `get` makes the trap easy to hit.

src/cli/plugins.go
```diff
@@ -218,6 +219,40 @@ pointing at Kestra's plugin registry to download them.`,
...
+		Args: cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			headers, _ := cmd.Root().PersistentFlags().GetStringArray(FlagHeader)
+			return runPluginsGet(cmd.OutOrStdout(), args[0], pluginsDir, forceRedownload, headers, mavenRepository, mavenUsername, mavenPassword, globalTimeout)
+		},
+		Annotations: map[string]string{AnnotationOffline: "true"},
+	}
+
+	cmd.Flags().StringVar(&pluginsDir, "plugins-dir", "./plugins", "Destination directory")
+	cmd.Flags().BoolVar(&forceRedownload, "force-redownload", false, "Re-download plugin if present and checksum-valid")
```
Malaydewangan09: The help says "checksum-valid" but there's no checksum anywhere in `downloadJAR`, the decision is a plain existence check. Suggest wording like "Re-download the plugin even if the file already exists" so it doesn't imply validation that isn't happening.

src/cli/plugins.go
```diff
@@ -218,6 +219,40 @@ pointing at Kestra's plugin registry to download them.`,
...
+	var pluginsDir string
+	var forceRedownload bool
+	var globalTimeout time.Duration
+	var mavenRepository string
+	var mavenUsername string
+	var mavenPassword string
+
+	cmd := &cobra.Command{
+		Use:   "get <groupId:artifactId:version>",
+		Short: "downloads a single plugin by Maven coordinates",
```
Malaydewangan09: Minor: `Short` reads better as a capitalized imperative to match the sibling commands, e.g. "Download a single plugin by Maven coordinates". Also "symbollic" is misspelled in the `validateCoordinateVersion` comment, and the `--global-timeout` help ("all downloads") was copied from the batch command though there is a single download here.

Malaydewangan09:
The red CI isn't from your code. Fork PRs don't get repo secrets, so the GCP auth step in e2e-tests runs with no credentials and fails.

src/cli/plugins.go
akashvijay98: @Malaydewangan09 Thank you for the review. I’ve added a validation check to parsePluginCoordinate to reject any parts containing /, \, or .. so that malinformed coordinates are rejected.

src/cli/plugins.go
akashvijay98: @Malaydewangan09 Thanks for pointing this. I have refactored downloadJAR() to use an atomic temp-file strategy. 
- Instead of writing directly to the final .jar file, I'm using `os.CreateTemp()` to create temporary file for the plugin `<plugin_name>.jar.part*`.
- Right after the transfer finishes, the code explicitly tells the system to close and save the file.  
- If the download gets interrupted by a timeout or a dropped connection, the code instantly deletes the temporary file.
- only when the plugin file is successfully downloaded and saved, we rename the file to the final plugin.jar file.

src/cli/plugins.go
akashvijay98: updated wording, removed checksum-valid. 
it now reads "Re-download the plugin even if it already exists"

Malaydewangan09:
I pushed some fixes and refactoring around counting the size of JARs and error messages.

A quick flag on issue #65: it says the download code already does SHA-1 checksum validation, but that code doesn't exist in the repo. The JAR is downloaded as-is, and "already up-to-date" just checks if the file exists.

Maven publishes a `.sha1` file for every artifact, so adding the check is small. Want to add it, or should I update the issue wording to match reality?

akashvijay98:
@Malaydewangan09 I'm working on the logic for adding checksum validation and will push the changes soon

akashvijay98:
@Malaydewangan09 I have added checksum validation. 
I've added `fetchExpectedSHA1()` which extracts the expected SHA-1 hash and `fileSHA1` which computes the checksum from a given file path.
Ive made changes in the `downloadJAR()` method where after downloading the JAR file, it verifies the new JAR against the expected SHA-1. If the JAR already exists locally, it compares its hash against the remote SHA-1. If they match, the download is skipped.

Malaydewangan09 (commented):
## Review: `plugins get` by Maven coordinates

Reviewed at head `99879ab`. Overall this is a solid, well-tested addition. The atomic temp-file download, path-traversal validation, version-alias rejection, and SHA-1 integrity checks are all real improvements over the previous `downloadJAR`. Test coverage is thorough (happy path, skip-existing, force, 404, invalid coordinate, timeout, 429 retry, basic auth, atomic cleanup, checksum mismatch on both existing and freshly downloaded files, and the missing-`.sha1` warn path).

A few items to address before merge.

### Should fix

**1. Files are not gofmt-clean.** `gofmt -l` flags both `src/cli/plugins.go` and `src/cli/plugins_test.go`. There is trailing whitespace on several lines in `plugins.go` (e.g. `636`, the doc comments at `738`/`754`/`755`, and `797`/`810`). Run `gofmt -w src/cli/` before merge, otherwise the format check fails in CI.

**2. Test error message typo.** In `TestRunPluginsInstall_429RetryThenSucceeds`:
```go
t.Errorf("expected 4s Maven attempts, got %d", attempts.Load())
```
Should read "expected 4 Maven attempts". The assertion itself (`!= 4`) is correct.

### Design / performance

**3. `fetchExpectedSHA1` runs before the exists-skip check in `downloadJAR`.** The checksum fetch is unconditional at the top of the function, ahead of the `!forceRedownload` + `os.Stat` skip. Two consequences:

- `plugins install` re-run when every JAR is already present used to issue ~0 Maven requests. It now issues one `.sha1` GET per plugin. For a full compatibility set (hundreds of plugins) that is a meaningful jump in request count and 429 exposure on every idempotent re-run.
- An existing, valid JAR can no longer be confirmed without network. If Maven is unreachable, `fetchExpectedSHA1` returns an error and the whole call fails instead of skipping the already-present file.

The verification itself is worth having. The cost of it is inherent when you want to validate an existing file against the remote checksum. Worth a conscious decision though: is per-file remote checksum validation on every `install` re-run the behavior you want, or should the skip stay a pure local fast-path with verification only on `get` / on download? At minimum, call the request-count change out so it is intentional.

### Robustness (low)

**4. The returned checksum is not validated as a 40-char hex string.** `fetchExpectedSHA1` returns `fields[0]` from a 200 response as-is. A proxy or repo returning a 200 HTML body would yield a bogus expected hash and produce a spurious "checksum mismatch" on every download. Consider validating length and hex, and treating anything else as warn-and-skip (same as the 404 path).

**5. Coordinate parts are only checked for `/`, `\`, `..`.** That covers path traversal well. It does not reject other URL-significant characters (`?`, `#`, space, `@`). With a fixed Maven Central base the blast radius is small, but a `?` in the version rewrites the request URL. A stricter allowed-character check on each part would close it.

**6. Non-404 checksum errors hard-fail the JAR download.** A `500` on the `.sha1` endpoint aborts a JAR that is otherwise downloadable. The 404 case already degrades gracefully with a warning. Consider extending that graceful degradation to other checksum-endpoint failures so JAR availability is not coupled to checksum-endpoint hiccups.

### Notes (non-blocking)

- Switching to `humanize.Bytes` also changes the existing `plugins install` output line from `%.1f MB` to humanized units. Human-facing only, so acceptable, but it is a visible change to an existing command.
- `plugins get` prints progress lines and does not support `--output json`. That matches its sibling `plugins install` (only `plugins list` calls `validateOutputFormat`), so this is consistent, just noting it against the "all commands support both outputs" convention.
- `--force-redownload` help text on `install` dropped "and pass checksum verification". Fine given the new flow, just confirm the wording still matches behavior.

Nice work on the atomic write and the checksum path. The main blockers are gofmt and the request-count behavior change on `install`.

akashvijay98:
@Malaydewangan09 Thanks for the detailed review. I have fixed the blocking issues and robustness. 
I decided to restore the pure local fast-path to avoid the 429 exposure and overhead on idempotent runs. 
We'll only fetch the remote .sha1 checksum when actually downloading a missing JAR or when --force-redownload is passed or else if the JAR exists on local we skip checksum validation.

Malaydewangan09 (approved):
## Re-review at `640943b`

Re-reviewed after the latest push. Every blocking and robustness item from my earlier review is addressed. Nice work.

**Resolved**
- gofmt is clean now.
- Test message typo fixed ("expected 4 Maven attempts").
- Idempotent fast-path restored: the `os.Stat` skip runs before any network call, so a re-run with everything present no longer fires a `.sha1` request per plugin. The checksum is fetched only when actually downloading.
- Expected checksum is validated as a 40-char hex string, and falls back to warn-and-skip on anything unexpected.
- Coordinate parts are now a strict allowlist (`^[A-Za-z0-9._-]+$`), which closes the URL-unsafe-character gap cleanly.
- A non-404 `.sha1` response no longer hard-fails the JAR download. It downgrades to a `[WARN]` and proceeds unverified.

**Minor, non-blocking (optional follow-ups)**
- The restored fast-path means existing files are skipped without integrity re-check. The checksum now guards fresh downloads only. Fine as a deliberate trade, just noting it.
- `plugins get` help still lacks an EE example. The only `Example` is the OSS `plugin-kafka`, so an EE user (`io.kestra.ee.*`) hits a 404 against the default Maven Central with no hint to pass `--maven-repository` + credentials. A one-line EE example would save that round trip.
- Prose nit in the `get` Long: missing a period after "...for a version" before "This lets users...".
- The old explicit `..` rejection is gone (the new regex allows dots). Not a real risk since the flat filename has no path separators and the URL stays same-host, just noting the guard relaxed.

Approving. The EE-help example is the only item worth doing before or shortly after merge, and it is optional.

akashvijay98:
Hi @jymaire I have fixed all the issues mentioned in the review, can you please review the changes?

fdelbrayelle (commented):
## Review at `640943b`

Verified: build, `go vet`, and `plugins` test suite all green; `gofmt -l` clean.

### New finding (design/perf, non-blocking)

**`downloadJAR` swallows `errRateLimited` from the checksum fetch instead of propagating it** (`src/cli/plugins.go:648-660`). When `fetchExpectedSHA1` exhausts its 429 retries, the error is treated like any other checksum failure — logged as a WARN and downgraded to "proceed without checksum" — rather than surfacing the rate-limit signal.

Under sustained 429s this means each plugin now burns a *full* retry cycle (default 45s: 5+10+30s) fetching the `.sha1`, falls through, then burns a *second* full retry cycle in the JAR-download loop before `errRateLimited` finally surfaces and stops the batch. So `plugins install` takes ~2x as long to detect and stop-early on sustained rate limiting compared to before checksum validation existed. `TestRunPluginsInstall_429ExhaustsRetriesStopsEarly` still passes (the JAR download's own 429 exhaustion still triggers stop-early), so this doesn't show up as a red test — it's just slower than it needs to be.

Suggest either fast-failing on `errRateLimited` specifically from the checksum step (skip the "proceed unverified" fallback for that one error), or sharing the retry budget between the checksum and JAR fetch instead of running two independent full cycles.

### Still open from the last approved review (optional, unchanged since this commit)

- No EE coordinate example in `get`'s help text (`plugins.go:243-247`) — an EE user hits a 404 against default Maven Central with no hint to use `--maven-repository` + credentials.
- Missing period: "...for a version\nThis lets users..." (`plugins.go:244`).

Nothing blocking — these were already called out as optional follow-ups.

akashvijay98:
Hi @fdelbrayelle, thanks for catching the issue!
I have added a small validation to fast fail on errRateLimited during the SHA-1 fetch so that if there is a 429 rate limiting error the function stops and returns immediately.
I've also added EE coordinate example and period to the get command's help text.

fdelbrayelle:
## Re-review at `8ad1c2b`

All three items from my previous review are addressed:

- **Rate-limit handling**: `downloadJAR` now fast-fails on `errRateLimited` from the SHA-1 fetch instead of downgrading to a WARN and running a second full retry cycle on the JAR download. Stop-early behavior under sustained 429s is restored.
- **EE example**: `plugins get` help now includes an `io.kestra.ee` coordinate example with `--maven-repository` and credentials.
- **Prose nit**: period added in the `Long` description.

Thanks @akashvijay98! LGTM once the `test` job is green.

fdelbrayelle (approved):

PR #98, opened 2026-08-11
feat(cli): warn once a day when a newer kestractl release is available

AcevedoR:
## What

`kestractl` warns on stderr when a newer release is available:

```
A new version of kestractl is available: v1.4.0 (you are on v1.2.0)
Download it from https://github.com/kestra-io/kestractl/releases/latest
```

## Throttling

A CLI has no session, so the equivalent is *the first command run after the cache goes stale*:

- State lives in `~/.kestractl/check_for_new_available_version.json` (`last_check` + `latest_version`), next to the existing `installation_id`.
- Under 24h old → notice rendered from cache, no network.
- Older/missing/corrupt/future-dated → one GitHub releases API call.

200 commands a day = 1 HTTP request.

## Asynchronous

The lookup never sits in front of the command. `startNewVersionCheck` (in `PersistentPreRunE`) prints any cached notice and hands a stale refresh to a goroutine; `awaitNewVersionCheck` (in `Execute`) collects it after the command has already done its work, waiting at most 500ms.

The throttle is stamped **before** the request, not after the response, which is what makes giving up free: an abandoned lookup still counts as today's check, so the next command is served from cache and the notice just appears a run later. The state file is written via temp file + rename, so a goroutine cut short at exit can't leave a half-written cache.

Measured against a local fake release server:

| | |
|---|---|
| responsive server, first run | **67ms** total, notice printed |
| cached run | **12ms** |
| server hanging 3s | **891ms** (grace-bounded), throttle still stamped |

## Safety

Notice goes to stderr (`--output json` stays clean), every failure path is silent, and the check is skipped for `dev` builds, on CI, and via `KESTRACTL_VERSION_CHECK_DISABLED=true`.

## Tests

19 unit tests in `src/cli/check_for_new_available_version_test.go`: cache hit asserts **zero** network calls, `startNewVersionCheck` asserts it returns before a 300ms fetch completes, the throttle is asserted to be stamped while the request is still in flight (server held open), grace expiry, no double-notice when a stale cache warned first, failed refresh keeps the cached version, all three skip paths, corrupt/future state, 0600 perms with no temp file left behind, and semver comparison.

`go test -race ./src/...` → 789 passing. `go vet` clean.

Also smoke-tested the real binary end to end: two runs, **one** server hit, `2>/dev/null` leaves only `kestractl v1.0.0` on stdout, and both skip env vars suppress it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,9 +119,9 @@ Set `KESTRACTL_TELEMETRY_DISABLED=true` to disable telemetry.
 
 ### Update Notifications
 
-The CLI warns on stderr when a newer release is available. The check for the latest release runs **at most once every 24 hours** — the result is cached in `~/.kestractl/update_check.json`, so only the first command you run after a full day makes a network call. Failures are silent and never block a command. The check is skipped automatically on CI and for `dev` builds.
+The CLI warns on stderr when a newer release is available. The check for the latest release runs **at most once every 24 hours** — the result is cached in `~/.kestractl/check_for_new_available_version.json`, so only the first command you run after a full day makes a network call. Failures are silent and never block a command. The check is skipped automatically on CI and for `dev` builds.
 
-Set `KESTRACTL_UPDATE_CHECK_DISABLED=true` to turn it off entirely.
+Set `KESTRACTL_VERSION_CHECK_DISABLED=true` to turn it off entirely.
 
 ## Usage
 
--- a/README.md
+++ b/README.md
@@ -119,7 +119,9 @@ Set `KESTRACTL_TELEMETRY_DISABLED=true` to disable telemetry.
 
 ### Update Notifications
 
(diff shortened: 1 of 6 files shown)


Malaydewangan09:
Nice, clean feature.

Two things I noticed, both minor:

1. The daily fetch runs synchronously in `PersistentPreRunE` before the actual command, so the first run each day can block the user's command for up to 2s if GitHub is slow. Bounded by the timeout + throttle so it's fine to ship, but might be nicer to run it in a goroutine and only print if it comes back in time.

2. A failed fetch caches `latest_version: ""` with a fresh `last_check`, so a single network blip means no update check for the next 24h (not just that run). Probably acceptable, just flagging the behavior.

Neither is a blocker. LGTM.

Malaydewangan09 (approved):

AcevedoR:
- I did the a-synchronous impl so the check doesnt slow down the first request
- "A failed fetch caches" is on purpose

PR #101, opened 2026-08-27
chore(sdk): bump go-sdk to v2.0.0-rc1

fdelbrayelle:
## Summary
- Pins `client-sdk/go-sdk` to the `go-sdk/v2.0.0-rc1` tag commit (Kestra 2.0 API). Upstream's `go.mod` still lacks a `/v2` suffix at this tag, so Go can't resolve `v2.0.0-rc1` as a proper semver import — pinned by commit pseudo-version instead. **This is an RC, not GA** — per `AGENTS.md`'s release checklist, this pin is not release-ready for a tagged `kestractl` release until go-sdk cuts a real `v2.x` tag (or adds the `/v2` module suffix).
- Adapts to v2's breaking API changes:
  - `superAdmin` → `instanceOwner` across users, service accounts, and invitations (fields, getters, and the dedicated patch endpoints — switched to the new `PatchServiceAccountInstanceOwner`/`PatchUserInstanceOwner`, since the old `*SuperAdmin` variants are explicitly deprecated in the SDK).
  - Dashboards CSV export: `ExportChartToCsv`/`ExportDashboardChartDataToCSV` removed in favor of `ExportChart`/`ExportDashboardChart` + a `format` query param.
  - `flows search-by-source`: `SourceSearchResult` dropped its `Model`/`Fragments` wrapper for flat `Namespace`/`Id`/`Editable`/`Matches` fields. **Output shape changes**: table/JSON output drops `REVISION`/`FRAGMENTS`, adds `EDITABLE`.
  - `namespaces export-plugin-defaults`/`import-plugin-defaults`: the dedicated raw-YAML endpoints are gone; plugin defaults are now just a field on the `Namespace` object. Reimplemented both commands as a namespace read (export) / read-modify-write (import). **Behavior change**: import no longer surfaces server-side validation warnings (the old endpoint could return them; the namespace PUT does not).
- Fixes a canary test (`TestConfirmArrayLabelsDecodeError`) tracking #83: v2's `FlowWithSource.Labels` is now a `MapObjectObject` accepting both array and map form, so the array-labels fixture decodes cleanly. The `tryParseFlow*FromError` fallbacks it guards are kept as a defensive fallback for other SDK type-mismatch-on-error cases, but are no longer exercised by that fixture — worth a follow-up cleanup PR once v2 is out of RC.

## Test plan
- [x] `go build -o kestractl`
- [x] `go test ./src/...` (all green, fixtures updated for renamed/removed endpoints)
- [ ] Manual smoke test against a live Kestra 2.0-rc instance (not available in this environment) — flag before merging to `main` for release.


fdelbrayelle:
Added \`v2.0\` to \`COMPATIBLE_KESTRA_VERSION.properties\` so e2e runs against it too.

⚠️ Unverified: this assumes a \`kestra-ee:v2.0-no-plugins\` image exists in the private registry (\`europe-west1-docker.pkg.dev/kestra-host/docker/kestra-ee\`). I have no way to check that from here — the e2e CI run on this PR is the real signal. If that tag doesn't exist yet, drop this entry until it does.

fdelbrayelle:
client-sdk pushed a real \`go-sdk/v2.0.0-rc1\` tag with the \`/v2\` module suffix fixed (run: https://github.com/kestra-io/client-sdk/actions/runs/33087121868).

Switched over: import path is now \`.../go-sdk/v2/kestra_api_client\` repo-wide, \`go.mod\` pins the real \`v2.0.0-rc1\` semver (no more commit pseudo-version). Same code as before, build+tests still green.

Still an \`-rc1\`, not GA — that's the only remaining release-checklist flag per \`AGENTS.md\`.

fdelbrayelle:
Reverted the \`v2.0\` e2e matrix entry — CI confirmed the earlier concern: \`kestra-ee:v2.0-no-plugins\` manifest not found in the registry. The go-sdk v2 tag and a Kestra server v2.0 image are separate releases; the server one doesn't exist yet. Back to \`v1.0\`–\`v1.3\`/\`develop\`, matching what's actually published. Will re-add once a real server image lands.

fdelbrayelle:
Fixed the tag: \`v2.0\` doesn't exist because Kestra server hasn't cut 2.0 GA — it's still on RCs. \`kestra-io/kestra\` tags show \`v2.0.0-rc1\` through \`v2.0.0-rc11\` right now. Swapped the matrix entry to \`v2.0.0-rc11\` (latest).

Still can't verify \`kestra-ee:v2.0.0-rc11-no-plugins\` exists in the private registry from here — CI run is the actual signal, same as before.

AcevedoR (approved):

PR #104, opened 2026-08-31
feat(install): default installs to latest v1, VERSION=<major> opt-in for v2

AcevedoR:
## Context

`main` is now the v2 line (go-sdk v2.0.0-rc2) while the v1 codebase lives on its own release branch. The public install command

```bash
curl -fsSL https://raw.githubusercontent.com/kestra-io/kestractl/main/install-scripts/install.sh | bash
```

must keep installing the **latest v1** release by default, with an explicit opt-in for v2. GitHub has exactly one repo-wide "latest" release and both lines tag into the same repo, so "latest" must be pinned per branch rather than left to release chronology.

## Changes

- **`.goreleaser.yml`**: `make_latest: "false"` — releases cut from `main` (v2) never claim GitHub's "latest" flag. The companion PR on the v1 branch pins `make_latest: "true"` so v1 releases always reclaim it. Flipping this to `"true"` is the deliberate one-line change for when v2 becomes the default channel.
- **`install-scripts/install.sh`**: a bare major `VERSION` (`VERSION=2` / `VERSION=v2`) resolves to the newest release of that major line via the releases-list API (prereleases included — useful while v2 only has rcs), then re-fetches that release by tag so the existing asset/checksum parsing is untouched. Default and exact-tag paths are unchanged.
- **`README.md`**: documents the `VERSION=2` opt-in.

## Resulting usage

| Command | Resolves to |
|---|---|
| `curl … \| bash` | latest v1 (unchanged) |
| `curl … \| VERSION=2 bash` | latest v2.x.x, prereleases included |
| `curl … \| VERSION=1.18.1 bash` | exact tag (unchanged) |

## Testing

Ran the modified script against the live GitHub API:
- default (no `VERSION`) → installed v1.18.1, checksum verified
- `VERSION=1` and `VERSION=v1` → resolved `v1.18.1`, installed
- `VERSION=2` → clean error `No release found for major version v2` (no v2 releases published yet)
- `VERSION=v1.17.0` → exact tag installed
- `bash -n` syntax check passes

## Related

Companion PR pins `make_latest: "true"` on the v1 maintenance branch (`releases/v1`).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -31,11 +31,6 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Trigger pkg.go.dev indexing
-        run: |
-          curl -sf "https://proxy.golang.org/github.com/kestra-io/kestractl/@v/${{ github.ref_name }}.info" > /dev/null
-          curl -sf "https://pkg.go.dev/github.com/kestra-io/kestractl@${{ github.ref_name }}" > /dev/null
-
       - name: Slack notification
         if: success()
         uses: kestra-io/actions/composite/slack-status@main
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -46,6 +46,10 @@ checksum:
 release:
   name_template: "{{ .Version }}"
   prerelease: auto
+  # main is the v2 line: never mark its releases as "latest" so the install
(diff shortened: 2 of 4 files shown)


AcevedoR:
Added `9a255d5`: removed the "Trigger pkg.go.dev indexing" step from the release workflow — its usefulness is unconfirmed, and pkg.go.dev indexes tagged versions on demand anyway.

fdelbrayelle (approved):

PR #105, opened 2026-08-31
chore(release): pin make_latest so v1 stays the default install channel

AcevedoR:
## Context

`releases/v1` is the new long-lived v1 maintenance branch (created at the `releases/v1.3.x` tip, one CI commit ahead of `v1.18.1`). `main` now carries the v2 line.

GitHub has exactly one repo-wide "latest" release, and both lines tag into the same repo — so whichever line's CI last claims `make_latest` wins. This must be pinned explicitly per branch, not left to release chronology.

## Change

`.goreleaser.yml`: `make_latest: "true"` — every v1 release explicitly reclaims GitHub's "latest" flag. The companion PR #104 on `main` pins `make_latest: "false"` for v2 releases, so the public install command (`curl … | bash`, which resolves `releases/latest`) keeps installing the latest v1 by default, while `VERSION=2` opts into v2.

## Verification (post-merge)

- After the next v1 tag push: `gh release view <tag> --json isLatest` → `true`
- After a v2 tag push from main: same command → `false`

## Related

- #104 (main side: `make_latest: "false"` + install script `VERSION=<major>` opt-in)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -36,11 +36,6 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Trigger pkg.go.dev indexing
-        run: |
-          curl -sf "https://proxy.golang.org/github.com/kestra-io/kestractl/@v/${{ github.ref_name }}.info" > /dev/null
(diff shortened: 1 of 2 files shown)


AcevedoR:
Added `d48a06d`: removed the "Trigger pkg.go.dev indexing" step from the release workflow — its usefulness is unconfirmed, and pkg.go.dev indexes tagged versions on demand anyway.

fdelbrayelle (approved):

PR #115, opened 2026-09-03
feat(ci): publish a multi-arch Docker image for every release

### LINKED ISSUE #97: Publish kestractl as a container image
The title really says it all.

## Use-case

I have two use-cases for this:

### 1. Automation from CI/CD platforms

We use GitLab to centrally manage the pushing of all deployable artifacts. For Kestra I ended up bucking that trend a bit, since at the time the whole "sync" capability only really worked in the way I expected using a `SyncFlows` and `SyncNamespaceFiles` inside a `ForEach`. This effectively turns a "push" into a "pull", which I was never very enthused with. kestractl seems like it's the path forward there, and if it was available in a container image it'd be trivial to write a GitLab CI/CD job that uses it to deploy things. I haven't worked much with other CI/CD platforms but I know they all can work nicely with tools distributed as images.

Regardless, I can always bake my own helper image internally, and I may end up doing that at some point in the future.

### 2. Developer experience

I have helper scripts to make it easy for devs to recursively sync up/down between their local Git clone of our flows and namespace files with their development instance. Right now this is accomplished through various commands available in the older Java-based CLI shipped in the Kestra binary. I prefer to use containers for these tasks to keep developer onboarding for our projects as simple as "install Docker then run scripts".

Additionally, most organizations don't like having people just pipe the output from a curl command through their shell, nor do they like downloading an unsigned binary off of GitHub. A container image isn't the only way to solve the distribution problem, but it would give me something I can name and point to on Docker Hub for governance purposes.

AcevedoR:
Closes #97.

kestractl is distributed as binaries only today, which makes it awkward to use from CI/CD platforms (the reporter uses GitLab) and gives organizations no nameable artifact to point governance at — they don't want to `curl | bash` or download an unsigned binary.

GoReleaser already produces static `CGO_ENABLED=0` linux amd64/arm64 binaries with version ldflags, so the images are thin COPY layers built from the same tag. No compilation inside Docker — same pattern as `kestra-io/kestra`'s Dockerfile. No Go code is touched.

## What gets published

Two variants, both multi-arch (`linux/amd64` + `linux/arm64`), both pushed to Docker Hub **and** GHCR:

| Tag | Contents |
| --- | --- |
| `2.0.0`, `2.0`, `2` | Alpine + `bash`/`git`/`curl`, non-root uid 1000 — usable as a CI job image |
| `2.0.0-static`, `2.0-static`, `2-static` | distroless `nonroot`, no shell — smallest surface |

Prereleases (`v2.0.0-rc13`) get their exact version tag only.

`dockers_v2` builds one real buildx manifest per variant, so no per-arch tags leak into the registries. The rolling tags are added afterwards with `regctl image copy`, which re-points the same manifest instead of rebuilding it — the convention already used in `kestra-io/actions`.

`DOCKER_LATEST` in `release.yml` is pinned per branch for exactly the same reason as `release.make_latest`: `latest` tracks v1 for as long as v1 is the default install channel. Documented in AGENTS.md next to the existing note.

## Also

A `docker-build` job is added to the Tests workflow. Without it the Docker path would only ever execute on a tag — it could only break in production.

Two implementation notes for reviewers:

- **`dockers_v2`, not `dockers` + `docker_manifests`** — `goreleaser check` flags the latter as deprecated and slated for removal in GoReleaser v3. `dockers_v2` also avoids the intermediate `-amd64`/`-arm64` tags entirely. It requires `COPY $TARGETPLATFORM/kestractl` in the Dockerfiles.
- **`--skip=publish` cannot be used for the dry run.** `dockers_v2` builds during GoReleaser's *publish* phase, so `--skip=publish` would skip the images silently and give a green job that tested nothing. Both the dry run and the CI job use `--snapshot --clean`.

## Verification

- `goreleaser check` clean, no deprecation warnings.
- Real `goreleaser release --snapshot --clean` locally: all four images build (17.6MB / 62.6MB Alpine amd64/arm64, 6.8MB / 27.3MB static).
- `kestractl version` reports the correct ldflags version/commit/date in both variants; `git`/`curl`/`bash` present in Alpine; processes run as uid 1000 and 65532 respectively.
- Config-file mount verified at **both** home paths (`/home/kestractl`, `/home/nonroot`) — `$HOME` resolves from `/etc/passwd` in both images.
- **Against a live Kestra `v2.0.0-rc13`**: both variants ran `namespaces list`, `flows validate ./flows` and `flows deploy ./flows` from a mounted directory — issue #97's exact use case — authenticating via `KESTRACTL_*` env vars and via a mounted `config.yaml`.
- `go test ./src/...` passes. (`gofmt -l` still flags `apps.go` and `test_suites.go`; both are pre-existing and untouched here.)

## Needs devops before the first tag

Pushing to Docker Hub requires the `kestra/kestractl` repository to exist and the org `DOCKERHUB_*` secrets to have push rights to it. I could not verify the secret names from here (`gh api orgs/kestra-io/actions/secrets` returns 403 without `admin:org`) — the OSS workflows in `kestra-io/actions` reference `DOCKERHUB_PASSWORD` while the EE ones reference `DOCKERHUB_TOKEN`, so this PR uses `${{ secrets.DOCKERHUB_PASSWORD || secrets.DOCKERHUB_TOKEN }}` to work with either.

Running the `Release` workflow via `workflow_dispatch` on this branch is a safe dry run: it exercises buildx and both registry logins and publishes nothing.

The GHCR half needs no new secrets, so it works today if Docker Hub provisioning lags.

`releases/v1` needs the same change cherry-picked with `DOCKER_LATEST: "true"` — separate PR, since per AGENTS.md v1 work targets `origin/releases/v1`.


AcevedoR:
Rebased onto `main` (`c09236d`), which brought in the auto-release flow from `cd68049`. That flow changes the assumptions this PR was written against, so three things needed fixing — each of which would have failed **silently or only on a real release**:

1. **`release.yml` now resolves the tag in one step and keys everything off that**, never off `github.event_name` or `github.ref`. On the `workflow_call` path both describe the *caller*: `github.ref` is `refs/heads/main`, and `github.event_name` is `workflow_dispatch` on a manual `auto-tag` re-run. The previous revision would therefore have (a) skipped the rolling tags on **every** auto-release and (b) turned a manual auto-tag re-run into a `--snapshot` build that published nothing while reporting success.

2. **`auto-tag.yml`'s `release` job now grants `packages: write`.** A reusable workflow inherits the caller's `GITHUB_TOKEN` and can only narrow it, so the grant inside `release.yml` does nothing on that path. This would have surfaced as a GHCR 403 on the first auto-release and nowhere earlier.

3. **A prerelease now moves the major-line tag `:2`.** Since `main` is auto-released as `v2.0.0-rc.N` and prereleases got no rolling tags, `:2`, `:2.0` and `:latest` would never have existed on the v2 line — and the README told people to `docker run kestra/kestractl:2`. `:2` now tracks the newest v2 release including rcs, which is exactly what `install.sh`'s `VERSION=2` default resolves to. Stable releases alone move `:X.Y` and `:latest`.

Also worth noting: `docker-build` is a job in `Tests`, and `auto-tag` fires on a green `Tests` run — so a broken image build now blocks the release rather than shipping a bad tag.

### Verification

Extracted the shell bodies from the workflow and ran them against every tag shape this repo produces:

| tag | rolling tags applied |
| --- | --- |
| `v2.0.0-rc.4` | `2`, `2-static` |
| `v2.0.0-rc1` (legacy undotted) | `2`, `2-static` |
| `v2.0.0` | `2.0`, `2` (+ `latest` when `DOCKER_LATEST=true`) |
| `v1.18.1` | `1.18`, `1`, `latest` |
| `nightly`, `v2.0` | none, exits 0 |

Confirmed `:latest` is never applied to a prerelease even with `DOCKER_LATEST=true`. Tag resolution checked across all four paths — `workflow_call` from `workflow_run`, `workflow_call` from a manual re-run, hand-cut tag push, and `workflow_dispatch` dry run.

`goreleaser check` clean; snapshot build produces all four images; both variants report correct ldflags version/commit/date; `go test ./src/...` passes. The live-Kestra checks from the original description still hold (unchanged code path).

Running the `Release` workflow via `workflow_dispatch` on this branch remains a safe dry run — it exercises buildx and both registry logins and publishes nothing.

jymaire:
## Code review

### High

**1. `.github/workflows/release.yml:98` — a Docker Hub push failure kills the GitHub release entirely, and the auto-release path can't retry.**

In GoReleaser's publish pipeline (`internal/pipe/publish/publish.go`) `dockerv2.Publish{}` runs *before* `release.Pipe{}`. So if the Docker Hub push fails — `kestra/kestractl` not created yet, or the org secrets lack push rights, both of which the PR body says are unverified — GoReleaser aborts before creating the GitHub release. `auto-tag.yml` has already pushed the tag, and its "tag already exists → `pushed=false`" guard means re-running Auto Tag skips it, so the tag exists with no release and `install.sh` (which reads `/releases`) resolves to the previous version. Recovery requires deleting the tag or running GoReleaser by hand. Suggest gating the Docker Hub half (or the whole `dockers_v2` publish) behind a flag until the registry is provisioned.

**2. `.github/workflows/release.yml:85` — the Docker Hub login step is unconditional and hard-fails when the secrets are absent.**

`docker/login-action@v3` errors with "Username and password required" when `username` is empty. `DOCKERHUB_USERNAME` is unverified per the PR body, and this step sits before `Run GoReleaser`, so a missing secret takes down *every* release, including the pure-binary part that works today — and it also fails the `workflow_dispatch` run the PR describes as "a safe dry run". Needs `if: ${{ secrets.DOCKERHUB_USERNAME != '' }}` or `continue-on-error`.

### Medium

**3. `.github/workflows/release.yml:164` — the success Slack notification is not gated on `dry_run`.**

`if: success()` with `suffix: ${{ steps.target.outputs.tag || github.ref_name }}`. A `workflow_dispatch` dry run on a branch produces `tag=""`, so the release channel gets a release-success message suffixed with the branch name (`emdash/docker-image-seiho`) for a release that published nothing. Add `if: success() && steps.target.outputs.dry_run == 'false'`.

**4. `.github/workflows/release.yml:21` — `workflow_dispatch` selected on a tag ref performs a real release, not a dry run.**

GitHub's dispatch UI lets you pick a tag as well as a branch. On a tag ref, `PUSHED_REF` is `refs/tags/v…`, so the resolver sets `dry_run=false` and the job pushes images and re-points the rolling tags for an already-released tag — then fails partway through when the GitHub release already exists. The comment says "run it on a branch", but nothing enforces it. Either reject a tag ref in the resolver, or add an explicit `dry_run` boolean input.

**5. `.github/workflows/release.yml:146` — the rolling tags are re-pointed with no version comparison, so an out-of-order release moves `:X`, `:X.Y` and `:latest` backwards.**

`regctl image copy` unconditionally re-points. A hand-cut patch on an older line (the normal mode of operation on `releases/v1`, whose tags are manual) — e.g. `v1.18.2` cut after `v1.19.0` — drags `:1` and, once `DOCKER_LATEST=true`, `:latest` back to the older minor. Same on `main` for any backport tag. A guard comparing the new triple against the tag currently behind `:$major` / `:latest` would prevent it.

**6. `.github/workflows/release.yml:117` — `regclient/actions/regctl-installer@main` is an unpinned third-party action in a job holding `contents: write`, `packages: write` and inherited Docker Hub credentials.**

Every other third-party action here is pinned (`docker/login-action@v3`, `actions/checkout@v4`); only first-party `kestra-io/actions` uses `@main`. A change to `regclient/actions`' default branch executes with the release token and registry logins already in the environment. Pin to a tag or SHA.

### Low

**7. `.goreleaser.yml:88` — OCI metadata is set via `labels`, which on a multi-platform build lands in the per-platform image configs, not the manifest index.**

GoReleaser only auto-scopes `annotations` to `index:` for multi-platform entries (`makeArgs`, `if len(d.Platforms) > 1`); `--label` flags are left untouched. `docker buildx imagetools inspect` on the published manifest list will show no `org.opencontainers.image.*`, which is the metadata GHCR uses to link a package to its repository. Mirror the same map into `annotations:`.

**8. `README.md:50` — the tag table documents v1 tags that do not exist and will not exist from this PR.**

`1`, `1.18`, `1.18.1` and `latest` are only produced once the same change is cherry-picked to `releases/v1` with `DOCKER_LATEST: "true"` (explicitly deferred to a separate PR). Until then `docker pull kestra/kestractl` (implicit `:latest`) fails, while the README tells the reader `latest` tracks stable v1. Worth softening the wording or landing the v1 cherry-pick first.

### Verified as correct

In case it saves the next reader the trip: the `$TARGETPLATFORM` context layout matches GoReleaser (`toPlatform` builds `<goos>/<goarch>`, artifacts copied to `filepath.Join(tmp, plat, art.Name)`); the snapshot per-arch tag suffix the CI smoke test assumes is real (`makeArgs`: `if len(d.Platforms) == 1 && ctx.Snapshot` appends `-` + `tagSuffix`, i.e. `-amd64`); `dist/metadata.json`'s `version` is the same `ctx.Version` that templates `{{ .Version }}`, so the two agree in snapshot mode; the `packages: write` claim about reusable workflows is accurate and the `auto-tag.yml` job-level grant is the right fix; and `dockers_v2`'s `ids`/`images`/`tags`/`labels` keys all match the current `config.DockerV2` struct.

AcevedoR:
Rebased onto `main` (`c0ed0fa`). The GA cutover invalidated several assumptions this PR was built on, so it's more than a replay — worth re-reading these bits rather than trusting the original description:

**`DOCKER_LATEST` flipped to `"true"`.** It was `"false"` with a comment saying to flip it "in the same change that flips `release.make_latest`" — `c0ed0fa` is that change. So v2 now claims the rolling `:latest` Docker tag, matching `make_latest: "true"`.

**One asymmetry that matters.** AGENTS.md now notes GitHub "cannot mark a draft or prerelease as latest, so `make_latest: \"true\"` is inert on any prerelease `main` might publish." Docker has no such protection — a registry will happily let `:latest` point at an rc. The guard is in `release.yml`: a prerelease moves `:X` only, never `:X.Y` or `:latest`. That rule was originally motivated by main being pre-GA; post-GA it's what keeps an rc from becoming what an untagged `docker pull` hands out, so it's still load-bearing, for a different reason. Rationale updated in AGENTS.md accordingly.

**The README tag table was rewritten.** It previously documented `2.0.0-rc.4`, `:2` tracking prereleases, and a v1 row (`1`, `1.18`, `latest`) with "`latest` stays on stable v1 until 2.0.0 is GA". All three are now wrong: v2 is GA, and per AGENTS.md `releases/v1` is legacy and not released any more, so **v1 image tags will never exist**. The table is now just the v2 line, in both variants. This also settles review finding #8 — there is no v1 cherry-pick to sequence, so nothing to soften.

### Re-verified after the rebase

Rolling tags under the new `DOCKER_LATEST=true`:

| tag | rolling tags |
| --- | --- |
| `v2.0.0` | `2.0`, `2`, `latest` (+ `-static` of each) |
| `v2.0.1` | `2.0`, `2`, `latest` |
| `v2.1.0` | `2.1`, `2`, `latest` |
| `v2.1.0-rc.1` | `2` only — no `2.1`, no `latest` |
| `nightly` | none, exits 0 |

`goreleaser check` clean; snapshot builds all four images and both variants report the correct ldflags version; `go test ./src/...` passes.

Still outstanding from your review, untouched: the unconditional Docker Hub login (#2 — note the suggested `if: ${{ secrets.… }}` can't work, `secrets` isn't available in a step `if`; needs mapping through `env`), the ungated success Slack (#3), the backwards-rolling-tag guard (#5), the unpinned `regclient/actions@main` (#6), and `labels:` without `annotations:` (#7). #1 and #4 are addressed by the retry commit.

jymaire (approved):

PR #116, opened 2026-09-03
feat(cli): best-effort Kestra 1.x compatibility for the v2 binary

AcevedoR:
## Why

We want **one kestractl binary**: focused on Kestra 2.x, but best-effort usable against Kestra 1.3 for the everyday commands (deploy/validate/get flows, run executions, nsfiles, kv, namespaces). The wire protocol barely moved between 1.3 and 2.0 (same `/api/v1/{tenant}` endpoints, no enum removed, flow YAML is passed through as raw source), so what actually broke was narrow, and all of it lives in how the generated Go SDK decodes 1.x responses and in 2.0-only request fields a 1.x server silently drops.

## Behaviour on a Kestra 1.3 instance with this PR

Everyday commands work as on 2.0: `flows deploy/validate/get/list/search/revisions/export/import/delete/disable/enable`, `executions run/get/list/latest/watch/kill/replay/flow-info`, `logs`, `nsfiles`, `kv`, `namespaces list/get/create/update/delete`, `tenants list/create/update`, `users`/`service-accounts`/`invitations` reads and non-superadmin writes, `flows usage-report`. Flow YAML is passed through as raw source, so 1.3 syntax (`pluginDefaults`, `json()`, …) deploys and validates according to the server you target.

Notable differences on 1.3:

- **2.0-only writes are refused instead of silently ignored.** `users/service-accounts/invitations … --superadmin`, `users/service-accounts set-super-admin`, and `--concurrency-limit/--concurrency-behavior/--quota` on `namespaces`/`tenants` exit with `<feature> is only available on Kestra 2.0 or later (the server runs 1.3.x)`. Nothing is sent to the server. `--superadmin=false` and the same commands without those flags work.
- **Superadmin status displays as `false` for everyone.** 1.3 reports it as `superAdmin`, the v2 model reads `instanceOwner`. Read-only cosmetic gap; managing superadmins on 1.3 needs kestractl v1.
- **Namespace `pluginDefaults` / `workerGroup` are not shown** by `namespaces get`, and there is no command to manage them (removed in v2, #103). `namespaces update` does not wipe them — verified: 1.3 keeps fields absent from the PUT.
- **`executions eval-expression` fails** (403): it already uses the 2.0 `/actions/eval` path; see #117.
- One extra `GET /api/v1/configs` per invocation, only on the guarded write commands and on flow/execution reads, to learn the server version.

## What

`src/cli/compat.go`:

**`compatTransport`** — one `http.RoundTripper` shared by both SDK clients (generated `client.API` and hand-written `client.Kestra`). On a 2xx JSON response from `/api/v1/{tenant}/flows/…` or `/api/v1/{tenant}/executions/…`, **and only when the server is known to run 1.x**, it adds `draft:false` to flow objects (`id`+`namespace`+`tasks`) and `tenantId:<tenant from the request path>` to execution objects (`flowId`+`flowRevision`+`state`). Details that matter:
- decodes with `UseNumber` so integers > 2^53 round-trip verbatim;
- cheap `bytes.Contains` pre-filter, so most responses are never decoded;
- non-JSON (SSE, zip), error statuses, other endpoints (a KV value can look exactly like a flow) and every 2.x response come back byte-identical — tested;
- tenant from the path, not the client, because `flows usage-report` walks several tenants.

**`requireKestra2(client, feature)`** — `Client.ServerVersion()` fetches `/api/v1/configs` once per process (`sync.Once`, reusing `fetchKestraVersion`). Guards placed exactly where a 1.x server would silently drop the field:

| Call site | Guarded when |
|---|---|
| `users create/update`, `service-accounts create`, `invitations create` | `--superadmin=true` |
| `users set-super-admin`, `service-accounts set-super-admin` | always |
| `namespaces create/update`, `tenants create/update` | `--concurrency-*` or `--quota` set |

Error: `--superadmin is only available on Kestra 2.0 or later (the server runs 1.3.35)`. An unreachable or unparseable version (develop builds) does not block — the guard turns silent no-ops into errors; it is not a gate on unknown servers.

`newTestClient` now mirrors production wiring, so the whole existing suite exercises the transport.

## Verified

- `go test ./src/...`, `-race`, `go vet`, `gofmt` on touched files.
- Live on **EE 1.3.35**: all four broken commands now pass; guards fire; `--superadmin=false`, plain `namespaces update`, `tenants update --name` still allowed; KV value shaped like a flow comes back untouched; multi-tenant `executions list --tenant acme` stamps `acme`; SSE `executions watch` unbuffered; `flows usage-report` unaffected.
- Live on **EE 2.0.0-rc13**: same commands, no behaviour change; shim provably inert.
- Independent Opus review + Opus QA pass; findings applied (large-integer corruption, whole-body walk scope, tenant source, guard consistency).

## Out of scope — pre-existing, filed separately

The QA pass surfaced bugs that predate this PR, filed separately:
- #117 execution action endpoints (`kill`, `replay`, `restart`, …) use 1.x paths and 403/404 on 2.0
- #118 `triggers list` renders empty rows on 2.0
- #119 `--verbose` prints the raw `Authorization` header on generated-SDK requests
- #120 default-context token overrides `--username/--password`
- #121 `kv get` NUMBER precision loss above 2^53
- #122 `logs list` renders LEVEL with literal quotes

### DIFF
--- a/src/cli/client.go
+++ b/src/cli/client.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net/url"
 	"strings"
+	"sync"
 
 	kestra "github.com/kestra-io/client-sdk/go-sdk/v2/kestra_api_client"
 	"github.com/spf13/viper"
@@ -23,6 +24,11 @@ type Client struct {
 	Kestra *kestra.KestraClient
 	Ctx    context.Context
 	Tenant string
+
+	// Kestra version reported by the server, resolved lazily by ServerVersion.
+	serverVersion     string
+	serverVersionErr  error
+	serverVersionOnce sync.Once
 }
 
 // newClientFunc is the client factory function. Override in tests.
@@ -67,6 +73,11 @@ func newClientDefault() (*Client, error) {
 	}
 	cfg.Debug = isVerbose
 
+	// One HTTP client for both SDK clients, so the Kestra 1.x response shim
+	// (see compat.go) applies to generated and hand-written endpoints alike.
+	httpClient, compat := newCompatHTTPClient()
+	cfg.HTTPClient = httpClient
+
 	parsed, err := parseHeaders(globalFlags.Headers)
 	if err != nil {
 		return nil, err
@@ -79,7 +90,7 @@ func newClientDefault() (*Client, error) {
 
 	// The hand-written client shares the same resolved host, headers, and debug
 	// setting; auth is appended per-branch below.
-	opts := []kestra.ClientOption{kestra.WithDebug(isVerbose)}
+	opts := []kestra.ClientOption{kestra.WithDebug(isVerbose), kestra.WithHTTPClient(httpClient)}
 	if len(parsed) > 0 {
 		opts = append(opts, kestra.WithHeaders(parsed))
 	}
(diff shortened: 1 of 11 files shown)


fdelbrayelle (approved):
@AcevedoR Don't forget to update the README accordingly for install.sh

PR #134, opened 2026-09-04
fix(nsfiles): delete a directory in one request so --recursive keeps siblings

### LINKED ISSUE #130: nsfiles delete --recursive on a sub-directory deletes sibling files (data loss)
## Summary
`kestractl nsfiles delete <ns> /dir/sub --recursive` removes **every file under `/dir`**, not just `/dir/sub`. Reproduced on Kestra EE **1.3.37** and **2.0.0-rc13/rc14** with kestractl `main` (`6bb00bc`, go-sdk v2.0.0-rc4).

## Repro
```
kestractl nsfiles upload qa.ns ./nsfiles /qa --no-root --override   # creates /qa/hello.txt and /qa/sub/nested.py
kestractl nsfiles list qa.ns --recursive                             # qa, qa/hello.txt, qa/sub, qa/sub/nested.py
kestractl nsfiles delete qa.ns /qa/sub --recursive                   # "2 path(s) deleted successfully"
kestractl nsfiles list qa.ns --recursive                             # EMPTY — hello.txt is gone
```

## Evidence
`--verbose` shows exactly two requests:
```
DELETE /api/v1/main/namespaces/qa.ns/files?path=qa%2Fsub%2Fnested.py   200
DELETE /api/v1/main/namespaces/qa.ns/files?path=qa%2Fsub               200
```
Raw `curl -X DELETE '…/files?path=qa%2Fsub'` (no leading slash) reproduces the parent wipe on 1.3.37; the same call with `path=%2Fqa%2Fsub` deletes only `sub` and keeps `hello.txt`.

## Likely cause
`normalizeNamespacePath` in `src/cli/nsfiles.go` strips the leading `/` before every request, and the server resolves the slash-less directory path against its parent. Sending `/`-prefixed paths for directory deletes avoids the wipe. The server-side behaviour for slash-less paths is worth a separate Kestra issue.

## Suggested e2e
`TestNsfilesDeleteRecursive_keepsSiblings` in `e2e_tests/nsfiles_usecases_test.go`: upload `/d/keep.txt` and `/d/sub/x.txt`, delete `/d/sub --recursive`, assert `nsfiles get /d/keep.txt` still succeeds.

Found by the `/kestractl-qa` matrix run of 2026-09-04 (T3.27).

AcevedoR:
## Summary

`kestractl nsfiles delete <ns> /dir/sub --recursive` deleted every file under `/dir`, not just `/dir/sub`. Reproduced locally on Kestra EE **2.0.0-rc13** and **1.3.35**.

## Root cause

Not the leading slash. The delete path enumerated the subtree and issued **one DELETE per path, deepest first**:

```
DELETE /files?path=qa%2Fsub%2Fnested.py   200
DELETE /files?path=qa%2Fsub               200   <- this one wipes /qa
```

Two server behaviours combine:

1. A directory is pruned as soon as its last child is removed. After the first DELETE, `qa/sub` no longer exists.
2. `DELETE /files?path=<missing path>` returns **200** and deletes the **parent** directory.

Verified with raw curl on both versions:

```
# /t130d contains keep.txt only
DELETE .../files?path=%2Ft130d%2Fnope   -> 200
GET    .../files/directory?path=%2Ft130d -> 404 "Directory not found: /t130d"
```

So the second DELETE — a well-formed request for a path the client had just caused to disappear — took the parent with it. A single DELETE on an existing directory was always correct and recursive:

```
# /t130b contains keep.txt and sub/x.txt
DELETE .../files?path=%2Ft130b%2Fsub    -> 200
GET    .../files/directory?path=%2Ft130b -> [keep.txt]   # sibling intact
```

## Fix

`runNamespaceFilesDelete` now issues **exactly one** DELETE, on the target path. The enumerated subtree is kept only to build the report, so `--output json` fields, the table rows, and the `N path(s) deleted successfully` line are unchanged.

`deleteNamespaceFilePath` also sends the path slash-prefixed, so the server resolves it absolutely. That is not what fixed this bug, but given that a DELETE which misses its target destroys the parent rather than failing, sending an unambiguous path is cheap insurance — and the issue reporter saw the slash-less form wipe the parent on 1.3.37.

Scope check: the only other `deleteNamespaceFilePath` caller (upload with `--override`) is already guarded by an existence check, so it can never issue a DELETE for a missing path.

## Tests

Unit (TDD — written first, confirmed failing with the pre-fix two slash-less DELETEs):

- `TestRunNamespaceFilesDelete_RecursiveIssuesOneDirectoryDelete` — asserts exactly one DELETE for `/qa/sub`, and that the report still names both paths.
- `TestRunNamespaceFilesDelete_SingleFileSendsSlashPrefixedPath`

```
go build -o kestractl .        PASS
go test ./src/...              PASS
go vet ./src/...               PASS
gofmt -l src/cli/nsfiles*.go   clean
```

(`gofmt -l src/` also flags `src/cli/apps.go` and `src/cli/test_suites.go`; both are pre-existing on `main` and untouched here.)

E2E — `TestNsfilesDeleteRecursive_keepsSiblings` in `e2e_tests/nsfiles_usecases_test.go`: uploads `<dir>/keep.txt` and `<dir>/sub/x.txt`, deletes `<dir>/sub --recursive`, asserts the nested file is gone and `keep.txt` still reads back.

```
cd e2e_tests && go vet ./...                          PASS
cd e2e_tests && go test ./... -run TestNsfiles        PASS  (vs local EE 2.0.0-rc13)
```

Also reproduced and re-verified by hand against **1.3.35** (the e2e suite hard-codes `localhost:9801`, so the 1.3 run was manual):

```
nsfiles delete system t130h/sub --recursive --verbose
  -> single DELETE ...files?path=%2Ft130h%2Fsub
nsfiles list system --path t130h --recursive
  -> t130h/keep.txt   # survived
```

Not verified: 1.3.37 and 2.0.0-rc14 exactly as reported in the issue — the local instances available were 1.3.35 and 2.0.0-rc13. The mechanism is identical on both of those, and the fix removes the offending request entirely rather than relying on a version-specific server behaviour.

## Note for the server team

Worth filing upstream independently of this client fix: `DELETE /api/v1/{tenant}/namespaces/{ns}/files?path=<missing>` returning 200 and deleting the parent directory is a data-loss primitive. Any client that races the empty-directory pruning can destroy an unrelated subtree. It should 404.

Fixes #130

### DIFF
--- a/src/cli/nsfiles.go
+++ b/src/cli/nsfiles.go
@@ -512,7 +512,6 @@ func runNamespaceFilesDelete(client *Client, namespace, targetPath string, recur
 	// the deepest-first walk reaches the directory itself that path no longer
 	// exists — and a DELETE on a path that does not exist deletes its *parent*,
 	// taking every sibling file with it.
-	failed := 0
 	deleteErr := deleteNamespaceFilePath(client, namespace, normalizedPath)
 	for _, target := range deleteTargets {
 		result := namespaceFileDeleteResult{
@@ -522,11 +521,17 @@ func runNamespaceFilesDelete(client *Client, namespace, targetPath string, recur
 		if deleteErr != nil {
 			result.Success = false
 			result.Error = deleteErr.Error()
-			failed++
 		} else {
 			result.Success = true
 		}
 		results = append(results, result)
+	}
+
+	// One request, so at most one failure — counting the reported rows instead
+	// would claim "6 error(s)" for a single 403 on a directory of five entries.
+	failed := 0
+	if deleteErr != nil {
+		failed = 1
 	}
 
 	if err := renderNamespaceFilesDeleteResults(results, renderer); err != nil {
--- a/src/cli/nsfiles.go
+++ b/src/cli/nsfiles.go
@@ -504,16 +504,24 @@ func runNamespaceFilesDelete(client *Client, namespace, targetPath string, recur
 		})
 	}
 
+	// One DELETE removes the whole subtree, so the enumerated targets are only
+	// there to report what went away.
+	//
+	// Deleting them one by one instead destroys data (issue #130): the server
+	// prunes a directory as soon as its last child is removed, so by the time
+	// the deepest-first walk reaches the directory itself that path no longer
+	// exists — and a DELETE on a path that does not exist deletes its *parent*,
+	// taking every sibling file with it.
 	failed := 0
+	deleteErr := deleteNamespaceFilePath(client, namespace, normalizedPath)
 	for _, target := range deleteTargets {
-		err := deleteNamespaceFilePath(client, namespace, target.Path)
 		result := namespaceFileDeleteResult{
 			Path: target.Path,
 			Type: target.Type,
 		}
-		if err != nil {
+		if deleteErr != nil {
 			result.Success = false
-			result.Error = err.Error()
+			result.Error = deleteErr.Error()
 			failed++
 		} else {
 			result.Success = true
@@ -731,8 +739,14 @@ func ensureNamespaceDirectories(client *Client, namespace, dirPath string) error
 	return nil
 }
 
+// deleteNamespaceFilePath removes a namespace file, or a directory and
+// everything under it, in a single request.
+//
+// The path is sent slash-prefixed so the server resolves it absolutely rather
+// than relative to anything: a DELETE that misses its target does not fail, it
+// deletes the parent directory instead (issue #130).
 func deleteNamespaceFilePath(client *Client, namespace, path string) error {
-	_, err := client.API.FilesAPI.DeleteFileDirectory(client.Ctx, namespace, client.Tenant).Path(path).Execute()
+	_, err := client.API.FilesAPI.DeleteFileDirectory(client.Ctx, namespace, client.Tenant).Path(absoluteNamespacePath(path)).Execute()
 	if err != nil {
 		return formatSDKError(err)
 	}
@@ -907,6 +921,15 @@ func normalizeNamespacePath(path string) string {
 		return ""
 	}
 	return trimmed
+}
+
+// absoluteNamespacePath turns a normalized (slash-less) namespace path into the
+// absolute form the API resolves unambiguously. An empty path stays empty.
+func absoluteNamespacePath(path string) string {
+	if path == "" || strings.HasPrefix(path, "/") {
+		return path
+	}
+	return "/" + path
 }
 
 func joinNamespacePath(base, child string) string {
--- a/src/cli/nsfiles.go
+++ b/src/cli/nsfiles.go
@@ -527,11 +527,12 @@ func runNamespaceFilesDelete(client *Client, namespace, targetPath string, recur
 		results = append(results, result)
 	}
 
-	// One request, so at most one failure — counting the reported rows instead
-	// would claim "6 error(s)" for a single 403 on a directory of five entries.
+	// The report is path-shaped: the one request removes or fails the whole
+	// subtree, so on failure every enumerated path is reported as not deleted
+	// and the error count matches the rows the report marks failed.
 	failed := 0
 	if deleteErr != nil {
-		failed = 1
+		failed = len(results)
 	}
 
 	if err := renderNamespaceFilesDeleteResults(results, renderer); err != nil {
--- a/e2e_tests/nsfiles_usecases_test.go
+++ b/e2e_tests/nsfiles_usecases_test.go
@@ -50,6 +50,38 @@ func TestNsfilesUpload_e2e(t *testing.T) {
 	require.Error(t, err, "file should have been deleted")
 }
 
+// Regression test for #130: deleting a subdirectory recursively used to wipe
+// every sibling file in the parent directory.
+func TestNsfilesDeleteRecursive_keepsSiblings(t *testing.T) {
+	dir := "e2e-nsfiles-tests/" + randomId()
+	keep := dir + "/keep.txt"
+	nested := dir + "/sub/x.txt"
+
+	_, stderr, err := RunAuthenticatedCliCmd(t, "nsfiles", "upload", "system", "./README.md", keep)
+	require.Empty(t, stderr)
+	require.NoError(t, err)
+
+	_, stderr, err = RunAuthenticatedCliCmd(t, "nsfiles", "upload", "system", "./README.md", nested)
+	require.Empty(t, stderr)
+	require.NoError(t, err)
+
+	t.Cleanup(func() {
+		_, _, _ = RunAuthenticatedCliCmd(t, "nsfiles", "delete", "system", dir, "--recursive", "--force")
+	})
+
+	_, stderr, err = RunAuthenticatedCliCmd(t, "nsfiles", "delete", "system", dir+"/sub", "--recursive")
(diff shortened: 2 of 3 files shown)


AcevedoR (commented):
Reviewed as PM.

- Root cause is convincingly evidenced (single-request directory delete is correct on 1.3 and 2.0; the second DELETE on the already-pruned path is what took the parent).
- Fix is minimal: one DELETE on the target, slash-prefixed as defence in depth. Report shape (table, JSON, summary line) unchanged.
- Requested one change during review (summary over-counted failures as one per enumerated row); addressed in 3ef933b with a failing-first unit test.
- Unit tests + full e2e matrix (v1.0–v1.3, 2.0.0-rc13) green.

Ready to merge from my side. Follow-up worth filing upstream on Kestra: DELETE on a non-existent namespace-file path returns 200 and removes the parent directory.

src/cli/nsfiles.go
```diff
@@ -504,23 +504,36 @@ func runNamespaceFilesDelete(client *Client, namespace, targetPath string, recur
...
 			result.Success = true
 		}
 		results = append(results, result)
 	}
 
+	// One request, so at most one failure — counting the reported rows instead
+	// would claim "6 error(s)" for a single 403 on a directory of five entries.
+	failed := 0
+	if deleteErr != nil {
+		failed = 1
```
jymaire: This `failed` count never reaches the output. `renderNamespaceFilesDeleteResults` (nsfiles.go:871) independently recomputes `Failed`/`Success` by counting rows in `results`, and every row above is stamped with `deleteErr`.

So `nsfiles delete ns qa/sub --recursive --output json` on a directory with two files plus a failing DELETE emits `{"total":3,"success":0,"failed":3,...}`, and the table footer prints `0 path(s) deleted successfully, 3 failed` — while the error returned to the user says `delete completed with 1 error(s)`. That's exactly the "would claim 6 error(s)" inconsistency this block set out to remove; it survives in the rendered output, and `TestRunNamespaceFilesDelete_RecursiveFailureCountsOnce` locks it in by asserting 3 failed rows.

Fix: pass the authoritative `failed` into the renderer (or have the renderer take it as a parameter) instead of letting it recount rows.

jymaire: Re-checked at `4bf8b9a`/`6be7556` — this is still open. `6be7556` fixed the **error string** but not the **rendered report**, so the inconsistency is still user-visible.

### Why it still has to change

`runNamespaceFilesDelete` now computes the authoritative count (nsfiles.go:533-536) and uses it for exactly one thing:

```go
failed := 0
if deleteErr != nil {
    failed = 1
}
// ...
return fmt.Errorf("delete completed with %d error(s)", failed)
```

But `renderNamespaceFilesDeleteResults` (nsfiles.go:871) recomputes it from the rows:

```go
failed := 0
for _, result := range results {
    if !result.Success {
        failed++
    }
}
```

and every row above is stamped with `deleteErr`. One DELETE, N reported rows — so a single 500 on a directory holding two files still prints:

```
{"total":3,"success":0,"failed":3,...}
0 path(s) deleted successfully, 3 failed
```

while stderr says `delete completed with 1 error(s)`. Same run, two different failure counts, and the report's is the wrong one: nothing failed three times, one request failed once.

That is the exact "would claim 6 error(s) for a single 403" problem the comment at nsfiles.go:530 says this block removes. It is removed from the returned error only; the JSON summary and the table footer — the parts a script or a human actually reads — still carry it. And `TestRunNamespaceFilesDelete_RecursiveFailureCountsOnce` (nsfiles_test.go:449-459) asserts three failed rows without ever asserting `summary.Failed`, so it locks the wrong number in rather than catching it.

### Proposed fix

Make the count an input to the renderer instead of something it re-derives. The row-level `Success`/`Error` stamping stays as-is — knowing which paths are gone is useful — it is only the *aggregate* that must come from the caller.

```go
-func renderNamespaceFilesDeleteResults(results []namespaceFileDeleteResult, renderer *Renderer) error {
-	failed := 0
-	for _, result := range results {
-		if !result.Success {
-			failed++
-		}
-	}
+// failed is the number of delete *requests* that failed, which is not the
+// number of failed rows: one DELETE removes a whole subtree, so a single
+// failure is reported against every enumerated path.
+func renderNamespaceFilesDeleteResults(results []namespaceFileDeleteResult, failed int, renderer *Renderer) error {
 	summary := namespaceFileDeleteSummary{
 		Total:   len(results),
 		Success: len(results) - failed,
 		Failed:  failed,
 		Results: results,
 	}
```

`Total`/`Success` need a decision too. With `failed` as a request count, `len(results) - failed` is meaningless (3 rows, 1 failed request -> `success: 2`, which is worse than what we have). Two coherent options:

1. **Request-shaped summary** — `Total: 1`, `Success: 1 - failed`, `Failed: failed`, and keep `Results` as the per-path detail. Honest, but changes the `total` field for the common success case from N to 1.
2. **Path-shaped summary, all-or-nothing** — `Total: len(results)`, and on failure `Success: 0`, `Failed: len(results)`, with the *returned error* also saying `len(results)`. Keeps the field meanings and makes both numbers agree, at the cost of the "6 error(s)" phrasing being back.

I'd take **(1)**: `total` is already 1 for the single-file and both early-return paths, the report's job is "what happened to the subtree" and `Results` covers that, and the summary's job is "did the operation succeed" — which is one request's outcome. It is a JSON field-value change on a not-yet-released fix, so no compat concern.

Then the three existing call sites:

- nsfiles.go:461 (path does not exist) -> `renderNamespaceFilesDeleteResults(results, 1, renderer)`
- nsfiles.go:477 (directory without `--recursive`) -> `renderNamespaceFilesDeleteResults(results, 1, renderer)`
- nsfiles.go:537 -> `renderNamespaceFilesDeleteResults(results, failed, renderer)`

both early returns already hardcode `1 error(s)`, so they line up.

And extend `TestRunNamespaceFilesDelete_RecursiveFailureCountsOnce` to assert the aggregate, not just the rows — this is the assertion whose absence let the bug through:

```go
if summary.Failed != 1 {
    t.Errorf("expected the summary to report 1 failed request, got %d: %s", summary.Failed, out.String())
}
if summary.Success != 0 {
    t.Errorf("expected no successes, got %d: %s", summary.Success, out.String())
}
```

Happy to push this if you'd rather I did it than round-trip.

AcevedoR: Thanks — you are right, the two numbers disagreed: the renderer recomputed `failed` by counting rows while the returned error hardcoded `failed = 1`, so a failed DELETE on a directory with two files printed `{"total":3,"success":0,"failed":3}` and then said `1 error(s)`.

Went with your option 2 (path-shaped, all-or-nothing) rather than option 1: the success case already reports `total: N` / "N path(s) deleted successfully", which is pre-existing released behaviour that option 1 would change. So on failure every enumerated path is reported as not deleted and the error now says N too. Renderer signature, JSON field set and the early-return paths are untouched.

Fixed in 3808d99. `TestRunNamespaceFilesDelete_RecursiveFailureCountsOnce` is renamed to `TestRunNamespaceFilesDelete_RecursiveFailureIsConsistentAcrossReportAndError` and now asserts the summary (`total 3 / success 0 / failed 3`) alongside `3 error(s)` in the error text; the success-path test pins `total 2 / success 2 / failed 0`.

jymaire (approved):

PR #136, opened 2026-09-04
fix(logs): send the level operation the server accepts on logs search

### LINKED ISSUE #132: Kestra 2.0: logs search --min-level sends filters[level][EQUALS], which the server rejects
## Summary
On Kestra **2.0.0-rc14**:
```
$ kestractl logs search -n qa.ns -q intentional --min-level ERROR
Error: API error: Invalid query filters: Provided query filters are invalid: Operation EQUALS is not supported for field LEVEL. Supported operations are GREATER_THAN_OR_EQUAL_TO, LESS_THAN_OR_EQUAL_TO, IN, NOT_IN
```
The identical command returns 3 rows on 1.3.37. Without `--min-level` the search works on both. kestractl `main` (`6bb00bc`, go-sdk v2.0.0-rc4).

## Evidence
```
GET /api/v1/main/logs/search?filters[level][EQUALS]=ERROR&filters[namespace][EQUALS]=qa.ns&filters[q][EQUALS]=intentional&page=1&size=50
```

## Fix
A "minimum level" is `GREATER_THAN_OR_EQUAL_TO`, which 2.0 accepts (verify it is also accepted by 1.3, otherwise gate on `serverEra`). Filter is built in `src/cli/logs.go` / `src/cli/query_filters.go`.

## Suggested e2e
`TestLogs_searchMinLevel` in a new `e2e_tests/logs_usecases_test.go`: run a flow with a WARN and an ERROR log, `logs search --min-level WARN` must return both and no INFO line, on every version in the matrix.

Found by the `/kestractl-qa` matrix run of 2026-09-04 (T4.11).

AcevedoR:
## Summary

`kestractl logs search --min-level ERROR` failed on Kestra 2.0 with:

```
Error: API error: Invalid query filters: Provided query filters are invalid: Operation EQUALS is not supported for field LEVEL. Supported operations are GREATER_THAN_OR_EQUAL_TO, LESS_THAN_OR_EQUAL_TO, IN, NOT_IN
```

The same command worked on 1.3, and dropping `--min-level` worked on both.

## Root cause

`buildLogSearchFilters` in `src/cli/logs.go` built every filter with `EQUALS`, including the level, so the request went out as
`filters[level][EQUALS]=ERROR`. Kestra 2.0 renamed the field to `LEVEL` and restricted its operations to
`GREATER_THAN_OR_EQUAL_TO`, `LESS_THAN_OR_EQUAL_TO`, `IN`, `NOT_IN`, rejecting `EQUALS` outright.

## Fix

The obvious swap to `GREATER_THAN_OR_EQUAL_TO` unconditionally would have broken 1.3, so the operation is now picked from the
existing `serverEra` probe (`Client.isLegacyServer`) in a new `minLevelSearchOp`:

- Kestra 2.0 and later, and unknown versions (develop builds, unreachable `/configs`) → `GREATER_THAN_OR_EQUAL_TO`
- Kestra 1.x → `EQUALS`, which is what its `MIN_LEVEL` field accepts

Defaulting an unknown server to the 2.0 operation matches the rest of the compat layer.

Only `logs search` was affected. Checked the other level-carrying commands:

- `logs list` / `logs download` / `logs follow` go through the SDK, whose `logExecutionFilters` already uses
  `OpGreaterThanOrEqualTo` for `FilterMinLevel`
- `logs delete` and `apps logs` still send the legacy scalar `minLevel` query param, not a filter

## 1.3 compatibility evidence

Verified live against two local Docker EE instances — 2.0.0-rc13 (`:9801`) and 1.3.35 (`:9802`). **1.3 rejects
`GREATER_THAN_OR_EQUAL_TO`**, which is why the fix is gated rather than a plain swap:

| server | `filters[level][EQUALS]=ERROR` | `filters[level][GREATER_THAN_OR_EQUAL_TO]=ERROR` |
| --- | --- | --- |
| 2.0.0-rc13 | 400 `Operation EQUALS is not supported for field LEVEL` | 200, rows returned |
| 1.3.35 | 200, rows returned | 400 `Operation GREATER_THAN_OR_EQUAL_TO is not supported for field MIN_LEVEL. Supported operations are EQUALS, NOT_EQUALS` |

`--min-level WARN` on 1.3.35 returns both `WARN` and `ERROR` rows, confirming its `MIN_LEVEL`/`EQUALS` really is minimum
semantics, so both server lines agree on the meaning of the flag and the e2e assertions hold on either.

With the patched binary, `logs search -n qa.kestractl --min-level ERROR` returns rows on **both** 9801 and 9802.

## Tests

Written failing-first: the new unit test reproduced the issue's exact query string
(`filters[level][EQUALS]=ERROR&filters[namespace][EQUALS]=qa.ns&filters[q][EQUALS]=intentional&page=1&size=50`) before the fix.

- `src/cli/logs_test.go`: `TestRunLogsSearch_MinLevelOperationPerServerEra` asserts the URL-decoded request query per era
  (2.0 → `GREATER_THAN_OR_EQUAL_TO`, 1.3 → `EQUALS`, unknown → `GREATER_THAN_OR_EQUAL_TO`), and that the rejected operation is
  absent in each case. `TestBuildLogSearchFilters` updated for the new signature and per-field operation.
- `e2e_tests/logs_usecases_test.go` (new): `TestLogs_searchMinLevel` deploys a flow logging one line per level with a unique
  marker, runs it with `--wait`, then asserts `logs search --min-level WARN` returns the WARN and ERROR lines and not the INFO
  one. Wrapped in `require.Eventually` because the log store is written asynchronously.

Commands run:

- `go build -o kestractl .` — ok
- `go test ./src/...` — ok
- `go vet ./src/...` — ok
- `gofmt -l src/cli/logs.go src/cli/logs_test.go` — clean (`src/cli/apps.go` and `src/cli/test_suites.go` are unformatted on
  `main` already and were left alone)
- `cd e2e_tests && go vet ./...` — ok
- `cd e2e_tests && go test -run TestLogs_searchMinLevel -v ./...` against 2.0.0-rc13 — **PASS (6.05s)**

Not verified: the e2e test was not executed against 1.3.35, because `e2e_tests/utils_test.go` hard-codes `:9801`. The 1.3 half
of the behaviour was verified with the patched CLI and raw API calls against the 1.3.35 instance instead (table above); CI's
version matrix will run it.

Fixes #132

### DIFF
--- a/e2e_tests/logs_usecases_test.go
+++ b/e2e_tests/logs_usecases_test.go
@@ -1,0 +1,94 @@
+package e2e_tests
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+const logsTestNamespace = "e2e.logs"
+
+// TestLogs_searchMinLevel covers #132: `logs search --min-level` sent
+// filters[level][EQUALS], which Kestra 2.0 rejects with
+// "Operation EQUALS is not supported for field LEVEL". The two server lines
+// accept different operations on that field and each 400s the other's, so this
+// runs on both halves of the version matrix and is the assertion that neither
+// regresses.
+//
+// The subject is a flow that logs one line per level, so "minimum level" has an
+// observable meaning: searching from WARN up must return the WARN and ERROR
+// lines and never the INFO one.
+func TestLogs_searchMinLevel(t *testing.T) {
+	marker := "kestractl-e2e-minlevel-" + randomId()
+	flowID := "e2e-logs-min-level-" + randomId()
+
+	deployLogLevelsFlow(t, flowID, marker)
+	runLogLevelsExecution(t, flowID)
+
+	// The log store is written asynchronously by the worker, so the search can
+	// legitimately be empty for a moment after the execution terminates.
+	var stdout string
+	require.Eventually(t, func() bool {
+		out, stderr, err := RunAuthenticatedCliCmd(t,
+			"logs", "search", "-n", logsTestNamespace, "-q", marker, "--min-level", "WARN")
+		if err != nil {
+			t.Logf("logs search failed, retrying\nstdout: %s\nstderr: %s", out, stderr)
+			return false
+		}
+		stdout = out
+		return strings.Contains(out, marker+"-warn") && strings.Contains(out, marker+"-error")
+	}, 60*time.Second, 2*time.Second, "expected the WARN and ERROR lines in the search output")
+
+	require.NotContains(t, stdout, marker+"-info",
+		"--min-level WARN must not return the INFO line:\n%s", stdout)
+	require.Contains(t, stdout, "WARN")
+	require.Contains(t, stdout, "ERROR")
+}
+
+// deployLogLevelsFlow deploys a flow that logs one message per level, each
+// tagged with marker so the search can find exactly these lines. It uses only
+// io.kestra.plugin.core.log.Log, which the plugin-less image this suite runs
+// against ships, and --override so a rerun against a persistent instance is
+// not a hard failure.
+func deployLogLevelsFlow(t *testing.T, flowID, marker string) {
+	t.Helper()
+
+	source := fmt.Sprintf(`id: %s
+namespace: %s
+tasks:
+  - id: log-info
+    type: io.kestra.plugin.core.log.Log
+    level: INFO
+    message: %s-info
+  - id: log-warn
+    type: io.kestra.plugin.core.log.Log
+    level: WARN
+    message: %s-warn
+  - id: log-error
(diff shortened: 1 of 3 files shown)


AcevedoR (commented):
Reviewed as PM.

- Good catch that a plain swap to GREATER_THAN_OR_EQUAL_TO breaks 1.x (MIN_LEVEL accepts only EQUALS/NOT_EQUALS). Gating on the existing serverEra is the right shape and the unknown-era default matches the rest of the compat layer.
- Scope check confirmed only `logs search` built the level filter by hand; the SDK-routed paths already used the right operation.
- Unit test pins the exact query string per era; e2e test runs on both halves of the matrix and asserts INFO is excluded.
- Unit + full e2e matrix green.

Ready to merge from my side.

fdelbrayelle (approved):

Issue #121, opened 2026-09-03
kv get loses precision on NUMBER values above 2^53


AcevedoR:
## What happens
```
kestractl kv set ns NUMBER big 9007199254740993
kestractl kv get ns big            → 9007199254740992
```
Raw `curl` on `/api/v1/main/namespaces/ns/kv/big` returns `9007199254740993`, so the server is fine; kestractl (or the SDK) decodes the number into a `float64` on the way to rendering. Same on 1.3.35 and 2.0.0-rc13.

Same class of problem likely applies to any JSON value with large integers rendered through `map[string]any` (e.g. `kv get` of a JSON value containing epoch nanos).

## Suggested fix
Decode with `json.Decoder.UseNumber()` where kestractl unmarshals into `any` before rendering (`kv.go` `tryParseKVTypedValueFromError`, renderer paths), or keep the raw text for NUMBER values.

Pradumnasaraf:
Working. QA done ✅

PR #137, opened 2026-09-04
fix(namespaces): decode inherited variables from the flat server shape

### LINKED ISSUE #128: namespaces inherited-variables fails to decode on Kestra 2.0 (flat map vs nested SDK type)
## What happens
On Kestra **2.0.0-rc13**, `kestractl namespaces inherited-variables <ns>` fails before rendering anything:

```
Error: json: cannot unmarshal number into Go value of type map[string]interface{}
```

## Why
The Go SDK types the inherited-variables response as `map[string]map[string]interface{}`, but the 2.0 server returns a flat `map[string]any` of variable name to value. Any scalar value (a number, a string) fails to decode into the inner map.

## Also
Once decoding works, the command renders values with `fmt.Sprintf("%v")` (`namespaces.go`, inherited-variables render), so a large integer would print as `1.72…e+18`. It should go through the same number-preserving path as `namespaces get` (see the #121 follow-up PR).

## Repro
Set any variable on a namespace on 2.0.0-rc13, then run `namespaces inherited-variables <ns>`.

Found while assessing the #121 follow-ups.

AcevedoR:
## Summary

`kestractl namespaces inherited-variables <ns>` failed before rendering anything on Kestra 2.0:

```
Error: json: cannot unmarshal number into Go value of type map[string]interface{}
```

The command now decodes the response shape the server actually sends, and renders values without losing precision. A second commit extracts the raw-request plumbing this needed into one helper shared with the pre-existing webhook path.

## Root cause

The generated Go SDK types `GET /api/v1/{tenant}/namespaces/{id}/inherited-variables` as `map[string]map[string]interface{}` (`ApiInheritedVariablesRequest.Execute`). Both server lines answer with the **already-merged flat map** of variable name to value, so the first scalar value fails the typed decode and the whole call errors out.

Secondary bug: once decoding worked, the render used `fmt.Sprintf("%v")`, so a variable above 2^53 printed as `1.725e+18` and a structured value printed in Go's `map[a:1]` syntax.

## Raw response shapes observed live

Namespaces `iv` (`big`, `name`) and `iv.compat` (`nested`, `flag`) created as parents, then `GET .../namespaces/iv.compat.test/inherited-variables`:

Kestra **2.0.0-rc13** (localhost:9801):
```json
{"big":1725000000000000001,"name":"hello","nested":{"a":1},"flag":true}
```

Kestra **1.3.35** (localhost:9802):
```json
{"name":"hello","big":1725000000000000001,"flag":true,"nested":{"a":1}}
```

Both are flat — **1.3 does not return the nested shape the SDK models**, so no version-conditional path is needed. (For contrast, the sibling `inherited-secrets` endpoint *is* genuinely nested — `{"iv":[],"iv.compat":[]}` — on both servers, and is untouched here.)

## Fix

`fix(namespaces): decode inherited variables from the flat server shape`

- `fetchInheritedVariables` in `src/cli/namespaces.go` performs the request directly instead of via the typed SDK method, mirroring the SDK's URL and reusing its configured host, HTTP client (so the `compat.go` 1.x shims still apply), default headers and context-based auth. Errors go through `formatErrorBody`, matching the existing raw-JSON precedent (`tryParse*FromError`).
- `parseInheritedVariables` decodes with `json.Decoder.UseNumber`, so integers keep their digits verbatim rather than round-tripping through `float64`. A response that still used the nested shape decodes here too instead of being rejected, though it is not the intended shape: each namespace would land in `KEY` with its variables JSON-encoded into `VALUE`. No supported server version returns it.
- Values render through the existing `stringify` helper (a `json.Number` is a `fmt.Stringer`, so its digits pass through exactly, and a map is JSON-encoded). This is the same number-preserving intent as the `#121` work; a small local decode is used rather than importing that PR's `decodeJSONPreservingNumbers`, which is not on `main` yet — no symbol clash either way.
- Rows are sorted by variable name (output was previously in Go map order, i.e. random) and `NAMESPACE` reports the namespace that was queried, since the server no longer attributes each variable to its defining namespace. Column names and JSON keys (`namespace`, `key`, `value`, all strings) are unchanged.

Live output on both servers after the fix:

```
NAMESPACE       KEY     VALUE
iv.compat.test  big     1725000000000000001
iv.compat.test  flag    true
iv.compat.test  name    hello
iv.compat.test  nested  {"a":1}

Total variables: 4
```

## Follow-up commit: one shared raw-request helper

`refactor(cli): share the raw request helper between webhook and inherited-variables`

The fix above initially repeated ~40 lines verbatim from `triggerWebhookDirect` in `src/cli/executions.go` (base URL resolution, context auth, default headers, `HTTPClient` fallback, body read, `formatErrorBody` on >= 400). Both callers now go through one helper in `client.go`:

```go
func (c *Client) doRawRequest(method string, segments ...string) ([]byte, error)
```

It builds `/api/v1/{tenant}/<segments>` with each segment `url.PathEscape`d, so a namespace or flow id containing a slash or space cannot break out of its position. The webhook path keeps its exact previous behaviour, including the `"status %d"` message its error text is built from; **its unit tests are untouched and still pass**. Net effect on the two callers: 58 lines removed from `executions.go`, 56 from `namespaces.go`.

New `TestClientDoRawRequest` covers the helper directly: tenant path building, segment escaping, `Authorization: Bearer` from the token context, basic auth from the context, the 404 body being formatted into the message, and the status fallback when the body carries no message.

## `--verbose` does cover the raw path

An earlier version of this description claimed `--verbose` misses requests issued through `doRawRequest`. **That is wrong on current `main`**, and the note (plus the matching "Known limitation" comment in `client.go`) has been removed. Verbose dumping now lives in `compatTransport` (`dumpRequest`/`dumpResponse`, with `writeMaskedHeaders` masking credential-bearing headers), and `doRawRequest` sends through `cfg.HTTPClient`, which *is* that transport — so a raw request gets exactly the same masked dump as any SDK call. The `-v` run quoted earlier predates that transport dump landing on `main`.

Covered by `TestDoRawRequest_VerboseDumpGoesThroughCompatTransport`: the raw inherited-variables request produces `> GET …/namespaces/my.namespace/inherited-variables`, `< 200 OK` and the response body in the dump, with the bearer token masked.

## Tests

Written test-first; the three decoding unit tests fail on `main` with the exact reported error before the fix.

| command | result |
|---|---|
| `go build -o kestractl` | ok |
| `go test ./src/...` | ok |
| `go test ./src/cli/... -run TestRunNamespacesInheritedVariables` | ok (5 tests) |
| `go test ./src/cli/... -run TestClientDoRawRequest` | ok (6 subtests) |
| `go test ./src/cli/... -run Webhook` | ok (6 tests, unchanged) |
| `go vet ./src/...` | clean |
| `gofmt -l src/cli` | only pre-existing `apps.go`, `test_suites.go` (untouched here) |
| `cd e2e_tests && go vet ./...` | clean |
| `cd e2e_tests && go test ./...` | full suite PASS (65.6s) against 2.0.0-rc13 |

New unit tests for the command: flat map with scalar/bool/object values, exact rendering of `1725000000000000001` (and no `e+18`), `--output json` row shape and ordering, empty `{}` response, and a 404 surfacing the server message.

New e2e test `TestNamespacesInheritedVariables_flatScalarsAndLargeIntegers` creates a parent namespace with a `--variables-file` (YAML so the large value is sent as a JSON number, not a string), a child namespace, asserts both table and JSON output, and deletes both in `t.Cleanup`.

### Not verified

- The e2e suite hard-codes `localhost:9801`, so the new e2e case ran only against 2.0.0-rc13. The 1.3.35 behaviour was verified by hand with the built binary and with `curl` (shapes above), not by the suite.
- There is no e2e coverage of the webhook command on either server, so the refactor of `triggerWebhookDirect` is covered by unit tests only (all passing, untouched).
- `go.mod`/`go.sum` untouched; no SDK bump.

Fixes #128

### DIFF
--- a/e2e_tests/namespaces_usecases_test.go
+++ b/e2e_tests/namespaces_usecases_test.go
@@ -2,8 +2,11 @@ package e2e_tests
 
 import (
 	"encoding/json"
-	"github.com/stretchr/testify/require"
+	"os"
+	"path/filepath"
 	"testing"
+
+	"github.com/stretchr/testify/require"
 )
 
 func TestNamespacesList_system(t *testing.T) {
@@ -28,4 +31,47 @@ func TestNamespacesList_system_json(t *testing.T) {
 	require.NoError(t, err, "json from --output json should be valid")
 	require.NotEmpty(t, parsedJson)
 	require.Equal(t, "system", parsedJson[0]["id"])
+}
+
+// Both Kestra 1.3 and 2.0 answer /inherited-variables with a flat
+// {variable: value} map, which the SDK's map-of-maps type cannot decode as soon
+// as one value is a scalar; the command used to fail outright on 2.0. A value
+// above 2^53 must also keep every digit rather than being rendered through
+// float64. See https://github.com/kestra-io/kestractl/issues/128.
+func TestNamespacesInheritedVariables_flatScalarsAndLargeIntegers(t *testing.T) {
+	parent := "e2e-inherited-vars-" + randomId()
+	child := parent + ".child"
+
+	variablesFile := filepath.Join(t.TempDir(), "variables.yml")
+	require.NoError(t, os.WriteFile(variablesFile, []byte("big: 1725000000000000001\nname: hello\nnested:\n  a: 1\n"), 0o600))
+
+	_, stderr, err := RunAuthenticatedCliCmd(t, "namespaces", "create", parent, "--variables-file", variablesFile)
+	require.NoError(t, err, "stderr: %s", stderr)
+	t.Cleanup(func() { _, _, _ = RunAuthenticatedCliCmd(t, "namespaces", "delete", parent) })
+
+	_, stderr, err = RunAuthenticatedCliCmd(t, "namespaces", "create", child)
+	require.NoError(t, err, "stderr: %s", stderr)
+	t.Cleanup(func() { _, _, _ = RunAuthenticatedCliCmd(t, "namespaces", "delete", child) })
+
+	stdout, stderr, err := RunAuthenticatedCliCmd(t, "namespaces", "inherited-variables", child)
+	require.NoError(t, err, "stderr: %s", stderr)
+	require.Contains(t, stdout, "hello")
+	require.Contains(t, stdout, "1725000000000000001", "a large integer must not be rendered through float64")
+	require.NotContains(t, stdout, "e+18")
+
+	stdout, stderr, err = RunAuthenticatedCliCmd(t, "namespaces", "inherited-variables", child, "--output", "json")
+	require.NoError(t, err, "stderr: %s", stderr)
+
+	var rows []map[string]interface{}
+	require.NoError(t, json.Unmarshal([]byte(stdout), &rows), "json from --output json should be valid")
+
+	values := map[string]interface{}{}
+	for _, row := range rows {
+		key, _ := row["key"].(string)
+		values[key] = row["value"]
+		require.Equal(t, child, row["namespace"])
+	}
+	require.Equal(t, "1725000000000000001", values["big"])
+	require.Equal(t, "hello", values["name"])
+	require.Equal(t, `{"a":1}`, values["nested"])
 }
--- a/src/cli/client.go
+++ b/src/cli/client.go
@@ -5,6 +5,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
+	"net/http"
 	"net/url"
 	"os"
 	"strings"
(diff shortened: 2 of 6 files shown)


AcevedoR (commented):
Reviewed as PM.

- Root cause confirmed on both 1.3 and 2.0: the endpoint returns a flat map, the SDK type is simply wrong. Raw decode with UseNumber is the right call and also closes the large-integer rendering gap from #121.
- Requested during review: the raw-request boilerplate was a verbatim copy of the webhook helper. Addressed in 22a8634 with a shared `Client.doRawRequest` (segments escaped centrally), webhook behaviour byte-identical, its tests untouched, plus 6 helper tests.
- Also confirmed the `-v` dump does not cover raw requests (pre-existing for webhook); documented rather than duplicated, since PR #127 moves dumping into the shared transport.
- Unit + full e2e matrix green.

Ready to merge from my side. Note for whoever merges: small textual rebase against #127 in client.go is likely; no logical conflict.

jymaire:
## Code review

Build, `go vet ./src/...` / `./e2e_tests/...`, and `go test ./src/...` (1147 pass) are clean. Core logic holds up: `doRawRequest` reproduces the previous `triggerWebhookDirect` behaviour byte-for-byte (same base-URL resolution, per-segment `url.PathEscape`, same context auth order, same `formatErrorBody`), the compat transport's execution-action rewrite cannot match `/namespaces/{id}/inherited-variables` or `executions/webhook/...` (`looksLikeExecutionID` needs ≥16 alphanumerics), `stringify` does hit the `fmt.Stringer` branch for `json.Number` so digits survive, and the sorting + `Total variables` accounting are right.

Four findings, all doc/UX-level rather than correctness:

**1. `src/cli/client.go:287` — the "Known limitation" comment is false on current `main`.**
It says raw requests don't show up in the `--verbose` dump, but verbose dumping now lives in `compatTransport` (`dumpRequest`/`dumpResponse`, wired by `newClientDefault`), and `doRawRequest` sends through `cfg.HTTPClient`, which *is* that transport. Verified with a throwaway test (`viper.Set(FlagVerbose, true)`): the raw request dumps `> GET http://…/api/v1/main/namespaces/x/inherited-variables` plus the full masked response. The PR description's "`--verbose` does not cover the raw path (correction)" section is wrong for the same reason — that `-v` run predates the transport dump landing on `main`. Worth removing, since a maintainer trusting the note would either add a duplicate caller-local dump or wrongly report that raw requests are invisible to `-v`.

**2. `src/cli/namespaces.go:623` — the stated reason for going raw names the wrong client.**
Only the *generated* client types the response as `map[string]map[string]interface{}`; the hand-written client already types it flat — `go-sdk/v2@v2.0.0-rc4/kestra_api_client/namespaces_api.go:53` returns `doJSON[map[string]interface{}](…, tenantPath(tenant, "namespaces", id, "inherited-variables"), …)`, reachable via `client.Kestra.Namespaces().InheritedVariables(...)`. Staying raw is still defensible — that path decodes into `interface{}` without `UseNumber`, so it would round `1725000000000000001` — but the comment should say *that* is the reason, otherwise the next reader concludes no SDK surface exists and repeats the raw pattern where the hand-written client would do.

**3. `src/cli/namespaces.go:637` — a non-JSON 2xx now surfaces a raw parser error.**
Behind an SSO/reverse proxy that answers an unauthenticated GET with a 200 login page, the old path went `formatSDKError` → `formatErrorBody`, whose `isHTML` branch prints "received HTML response instead of JSON. Check your host URL and authentication". Now the user gets `failed to parse inherited variables response: invalid character '<' looking for beginning of value`. Routing the decode failure through `formatErrorBody(body, …)` restores the actionable message.

**4. `src/cli/namespaces.go:592` — the nested-shape fallback is described as "lossless" but mislabels rows.**
If any server/version returns `{"parent.ns":{"big":1}}`, the output is `NAMESPACE=<queried ns>`, `KEY=parent.ns`, `VALUE={"big":1}` — the KEY column holds a namespace and the variable names disappear into the value, silently, with no warning. Related (and acknowledged in the PR body): `NAMESPACE` is now the queried namespace on every row instead of the defining namespace, so `kestractl namespaces inherited-variables child` reads as if every variable were defined on `child`. A one-line hint in the table, or dropping/renaming the now-constant column, would keep the output honest.

AcevedoR:
Thanks @jymaire — all four are addressed in d37ac83.

1. **`client.go` "Known limitation" comment** — you are right, it is false on current `main`. Deleted; the doc comment now says the request goes through the shared compat transport, so it gets the same `--verbose` masked dump as an SDK call. Verified with `TestDoRawRequest_VerboseDumpGoesThroughCompatTransport`, which asserts the raw inherited-variables request dumps `> GET …/inherited-variables`, `< 200 OK` and the body, with the bearer token masked. The matching section in the PR description is rewritten too.

2. **Wrong client named as the reason for going raw** — reworded. The comment now says only the *generated* client mistypes the shape as `map[string]map[string]interface{}`, while the hand-written `Namespaces().InheritedVariables` already returns a flat `map[string]interface{}` but decodes with a plain `json.Unmarshal`, so without `UseNumber` a value above 2^53 rounds. That precision loss is now stated as the actual reason for reading the body ourselves. Raw path kept.

3. **Non-JSON 2xx surfacing a raw parser error** — fixed, and this is the only behaviour change. `parseInheritedVariables` now routes an HTML body through `formatErrorBody(body, "")`, so an SSO/proxy login page again produces `received HTML response instead of JSON. Check your host URL and authentication`; any other unparseable body keeps the decode error behind the same `failed to parse inherited variables response` prefix. Written test-first: the HTML test fails on the previous commit with `invalid character '<' looking for beginning of value`.

4. **Constant `NAMESPACE` column and the "lossless" claim** — decision is to keep the `namespace` JSON field and the NAMESPACE column for output-shape stability, and to document the truth instead of hinting at it in the table. The command's `Long` help now says the server merges inheritance before answering, so the namespace shown is the one queried, not the one that defined each variable — that attribution is simply not in the response on either 1.3 or 2.0. The nested-shape comment no longer says "lossless": it now states that such a body would render each namespace as a KEY with its variables JSON-encoded in VALUE, which is readable but not the intended shape, and that no supported server version returns it.

`go build`, `go test ./src/...`, `gofmt -l src/cli` (only the pre-existing `apps.go`/`test_suites.go`), `go vet ./src/...` and `cd e2e_tests && go vet ./...` are clean.

jymaire (approved):

Issue #118, opened 2026-09-03
triggers list renders empty rows on Kestra 2.0 (nested trigger/state body)


AcevedoR:
## What happens
On Kestra **2.0.0-rc13**, `kestractl triggers list` prints rows with blank NAMESPACE / FLOW / TRIGGER ID / TYPE; `-o json` shows `""` for all four. On 1.3.35 the same command renders correctly.

## Why
2.0 returns a nested body:

```json
{"results":[{"trigger":{"id":"…","type":"…","cron":"…"},"state":{"namespace":"…","flowId":"…","triggerId":"…","updatedAt":"…","nextEvaluationDate":"…","disabled":false}}]}
```

kestractl reads the flat top-level fields of the 1.3 shape. The trigger rendering was never migrated to the v2 model (`ApiTriggerAndState` → `trigger` + `state`).

## Repro
Deploy a flow with a `Schedule` trigger on rc13, then `triggers list`.

Found during QA of the 1.x compatibility PR.

Pradumnasaraf:
Working. QA done ✅

PR #138, opened 2026-09-04
test(e2e): cover the trigger decode fix on both server lines

### LINKED ISSUE #131: Kestra 2.0: triggers search-for-flow and backfill-pause/unpause/delete fail to decode the response ("no value given for required property date")
## Summary
On Kestra **2.0.0-rc14** the server answers `HTTP 200` but kestractl exits 1:
```
$ kestractl triggers search-for-flow qa.ns qa-scheduled
Error: API error: no value given for required property date
```
Same for `triggers backfill-pause`, `backfill-unpause`, `backfill-delete`. All four work on 1.3.37. kestractl `main` (`6bb00bc`, go-sdk v2.0.0-rc4).

## Evidence
```
GET /api/v1/main/triggers/qa.ns/qa-scheduled?page=1&size=50 HTTP/1.1
HTTP/1.1 200 OK
Error: API error: no value given for required property date
```
Raw 2.0 response item:
```json
{"trigger":{"id":"every-minute","type":"io.kestra.plugin.core.trigger.Schedule","disabled":true,"cron":"* * * * *"},
 "state":{"namespace":"qa.ns","flowId":"qa-scheduled","triggerId":"every-minute","updatedAt":"…","nextEvaluationDate":"…","disabled":true,"locked":false,"kind":"SCHEDULE"}}
```
No top-level `date`; 2.0 nests everything under `trigger` / `state`. This is the same shape change that makes `triggers list` render blank rows on 2.0 (fields read as flat 1.3 attributes).

## Likely cause
The SDK `Trigger` model still requires a flat `date` property. Needs a compat shim in `src/cli/compat.go` next to `fillCompatDefaults` (unwrap `trigger`/`state`, or relax `required` in the client-sdk spec post-processing).

## Suggested e2e
`TestTriggers_searchForFlow_decodes` gated with `skipBelowKestra(2,0)`, deploying a flow with a disabled Schedule trigger.

Found by the `/kestractl-qa` matrix run of 2026-09-04 (T4.2, T4.6).

AcevedoR:
## Summary

Issue #131 needs no new code: **the fix already sits in open PR #127**, which rewrites exactly the two functions the issue names. What was missing is e2e coverage, so this PR is stacked on `emdash/bug-fix-sawvr` and adds `e2e_tests/triggers_usecases_test.go` — three tests that run the real binary against whichever Kestra the suite is pointed at, so both server lines in `COMPATIBLE_KESTRA_VERSION.properties` assert the decode.

> Base is #127's branch, not `main`, deliberately: on `main` these tests fail (that is the bug), so a PR against `main` would be red until #127 lands. Merge #127 first, then this.

## Root cause

Kestra 2.0 reshaped every `/triggers/**` response, and the generated Go SDK still ships the 1.x models for those paths. `runTriggersSearchForFlow` and `runTriggersBackfillOp` decoded into `kestra.Trigger`, whose `date` is required and which a 2.0 server never sends — so the SDK rejected an HTTP 200 body and the command exited 1.

Raw `GET /api/v1/main/triggers/{namespace}/{flowId}?page=1&size=50`, captured live:

**Kestra 1.3.35** — flat `Trigger`, carries `date`:
```json
{"results":[{"tenantId":"main","namespace":"bug131","flowId":"sched131","triggerId":"every_min",
  "date":"2026-09-04T15:10:10Z","nextExecutionDate":"2026-09-04T15:15:00Z","disabled":false}],"total":1}
```

**Kestra 2.0.0-rc13** — `ApiTriggerState`, no `date` at all:
```json
{"results":[{"namespace":"bug131","flowId":"sched131","triggerId":"every_min",
  "updatedAt":"2026-09-04T15:08:55Z","nextEvaluationDate":"2026-09-04T15:10:00Z",
  "disabled":true,"locked":false,"kind":"SCHEDULE"}],"total":1}
```

The paged `/triggers/search` used by `triggers list` changed the same way, nesting the two halves under `trigger` and `state` — that is #118, the blank-row symptom of the same shape change.

## Fix

None here. #127's `src/cli/compat_triggers.go` dispatches on server era and funnels both shapes into era-agnostic values (`fetchFlowTriggerRefs`, `backfillTriggerRef`), reading the 2.x bodies with the hand-written client that models them and the 1.x bodies raw. Adding a shim in `fillCompatDefaults` instead was considered and rejected for the same reason #127 documents: the transport would have to invent an `updatedAt` a 1.3 server legitimately omits.

## Relationship to PR #127

- #127 fixes #131 in full — its body already calls out `search-for-flow`, `unlock` and the single backfill ops, and its unit tests (`TestFetchFlowTriggerRefs_BothEras`, `TestBackfillTriggerRef_BothEras`, plus the dateless-1.3 regression bodies) cover every shape.
- This PR duplicates none of it: it touches only `e2e_tests/`, a module #127 does not modify, so there is no conflict surface.
- #131 should be closed by whichever of the two merges last; `Fixes #131` is on this one because it is the later half.

## Tests

Local Docker EE instances: **2.0.0-rc13** on `:9801`, **1.3.35** on `:9802`. A flow with one disabled `Schedule` trigger was deployed to both.

Reproduction on `main` (`6bb00bc`), against 2.0.0-rc13 — all four commands from the issue:
```
=== search-for-flow    Error: API error: no value given for required property date
=== backfill-pause     Error: API error: no value given for required property date
=== backfill-unpause   Error: API error: no value given for required property date
=== backfill-delete    Error: API error: no value given for required property date
=== triggers list      rows render blank (#118)
```

New e2e tests, `cd e2e_tests && go test -run 'TestTriggers_' -count=1 ./...`:

| code | server | result |
|---|---|---|
| `main` | 2.0.0-rc13 | **FAIL** ×3 — the reported error, verbatim |
| this branch (#127 + tests) | 2.0.0-rc13 | **PASS** — `ok e2e_tests 10.475s` |
| this branch (#127 + tests) | 1.3.35 | **PASS** — `ok e2e_tests 3.685s` |

Also on this branch: `go build`, `go test ./src/...` (ok), `go vet ./src/...`, `cd e2e_tests && go vet ./...`, `gofmt -l` clean on the new file.

### Not verified
- CI's own matrix (only rc13 and 1.3.35 were reachable locally); the version the issue reports, 2.0.0-rc14, and 1.3.37 were not.
- `triggers backfill-*` against a trigger that actually **has** a backfill. On both servers the trigger under test had none — 1.3 answers 404 for that, 2.0 accepts the call — which is why those subtests assert only that the response is readable, not a success message.
- The full e2e suite was not re-run end to end; only `-run TestTriggers_`.

Fixes #131

### DIFF
--- a/e2e_tests/triggers_usecases_test.go
+++ b/e2e_tests/triggers_usecases_test.go
@@ -1,0 +1,143 @@
+package e2e_tests
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+// Coverage for the trigger read commands across the version matrix.
+//
+// Kestra 2.0 changed the shape of every /triggers/** response: a per-flow
+// search that used to answer with a flat `Trigger` (namespace/flowId/triggerId/
+// date/nextExecutionDate) now answers with an `ApiTriggerState` (updatedAt/
+// nextEvaluationDate, no `date` at all), and the paged search nests the two
+// halves under `trigger` and `state`. The generated Go SDK still marks the 1.x
+// `date` required, so reading a 2.0 body through it failed the whole command
+// with "no value given for required property date" — issue #131 for
+// search-for-flow and the single backfill ops, #118 for `triggers list`.
+//
+// The unit tests pin each shape against a fake server; these run the real
+// binary against every version in COMPATIBLE_KESTRA_VERSION.properties, which
+// is the gap that let the decode failure reach a release unnoticed.
+
+const triggersTestNamespace = "e2e.triggers"
+
+// triggersTestTriggerID is the trigger id deployScheduledTestFlow declares.
+const triggersTestTriggerID = "every_min"
+
+// triggerDecodeFailure is the error a trigger response that the SDK model
+// rejects produces. It can never be a legitimate answer, on either server line,
+// so its absence is the assertion these tests are built around.
+const triggerDecodeFailure = "no value given for required property"
+
+// deployScheduledTestFlow deploys a flow carrying one disabled Schedule
+// trigger, and returns its flow id.
+//
+// The trigger is disabled so a run of this suite does not leave a schedule
+// firing every few minutes behind on a persistent instance; a disabled trigger
+// still gets a trigger state row, which is what these commands read.
+func deployScheduledTestFlow(t *testing.T, flowID string) string {
+	t.Helper()
+
+	source := fmt.Sprintf(`id: %s
+namespace: %s
+triggers:
+  - id: %s
+    type: io.kestra.plugin.core.trigger.Schedule
+    cron: "*/5 * * * *"
+    disabled: true
+tasks:
+  - id: hello
+    type: io.kestra.plugin.core.log.Log
+    message: kestractl e2e
+`, flowID, triggersTestNamespace, triggersTestTriggerID)
+
+	path := filepath.Join(t.TempDir(), flowID+".yml")
+	require.NoError(t, os.WriteFile(path, []byte(source), 0o600))
+
+	stdout, stderr, err := RunAuthenticatedCliCmd(t, "flows", "deploy", path, "--override")
+	require.NoError(t, err, "deploy failed\nstdout: %s\nstderr: %s", stdout, stderr)
+	require.Contains(t, stdout, "1 flow(s) deployed successfully")
+
+	return flowID
+}
+
+// eventuallyTriggerListed runs the given command until its output contains
+// want, and returns that output.
+//
+// The wait is for the server, not the CLI: a trigger state row appears only
+// once the scheduler has registered the freshly deployed flow, which on a 1.x
+// instance takes a few seconds. A decode failure is not waited out — it is a
+// terminal answer, and the whole point of these tests — so the loop fails on it
+// immediately rather than retrying until the deadline.
+func eventuallyTriggerListed(t *testing.T, want string, args ...string) string {
+	t.Helper()
+
+	deadline := time.Now().Add(60 * time.Second)
+	var stdout, stderr string
+	var err error
+	for {
+		stdout, stderr, err = RunAuthenticatedCliCmd(t, args...)
+		require.NotContains(t, stdout+stderr, triggerDecodeFailure,
+			"the trigger response was rejected at decode\nstdout: %s\nstderr: %s", stdout, stderr)
+
+		if err == nil && strings.Contains(stdout, want) {
+			return stdout
+		}
+		if time.Now().After(deadline) {
+			t.Fatalf("%q never listed %q within the deadline: %v\nstdout: %s\nstderr: %s",
+				strings.Join(args, " "), want, err, stdout, stderr)
+		}
+		time.Sleep(2 * time.Second)
+	}
+}
+
+// TestTriggers_searchForFlow_decodes covers issue #131: on Kestra 2.0 the server
+// answers 200 and kestractl exited 1 on the response body alone.
+func TestTriggers_searchForFlow_decodes(t *testing.T) {
+	flowID := deployScheduledTestFlow(t, "e2e-triggers-search")
+
+	stdout := eventuallyTriggerListed(t, flowID,
+		"triggers", "search-for-flow", triggersTestNamespace, flowID)
+
+	require.Contains(t, stdout, triggersTestTriggerID, "the trigger id is missing from the output")
+	require.Contains(t, stdout, triggersTestNamespace, "the namespace is missing from the output")
+}
+
+// TestTriggers_backfillOps_decode covers the other half of issue #131. Whether
+// a backfill exists to act on differs by server line — 1.x answers 404 for a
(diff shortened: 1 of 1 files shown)


AcevedoR (commented):
Reviewed as PM.

- Agree with the call: the code fix for #131 already lives in #127 (`compat_triggers.go` rewrites exactly `fetchFlowTriggerRefs` / `backfillTriggerRef`), so a second fix would only have conflicted. This PR adds the missing e2e coverage for search-for-flow, the three backfill ops, and `triggers list` rows, across the whole matrix.
- Tests treat a decode failure as terminal instead of waiting it out, and tolerate the legitimate 1.x 404 on backfill ops. Sensible.
- Full e2e matrix green on top of #127.

Merge order: #127 first, then this one (it is based on #127's branch; retarget to main after #127 lands). Ready from my side.

e2e_tests/triggers_usecases_test.go
```diff
@@ -0,0 +1,143 @@
...
+		})
+	}
+}
+
+// TestTriggers_list_rendersRows covers issue #118: the 2.0 {trigger, state}
+// body decoded into empty fields, so every row rendered blank.
+func TestTriggers_list_rendersRows(t *testing.T) {
+	flowID := deployScheduledTestFlow(t, "e2e-triggers-list")
+
+	stdout := eventuallyTriggerListed(t, flowID, "triggers", "list", "--size", "200")
```
jymaire: `triggers list --size 200` pages over **all** triggers in the tenant, so this asserts that a global page 1 happens to contain the trigger this test just deployed. On a long-lived instance — the docker-setup README documents reusing one across runs — with more than 200 trigger rows, the new trigger simply is not on page 1, and the test then fails after the 60s deadline with "the trigger id is missing from the listing", pointing at a render regression that does not exist.

Scoping the listing to this flow (`search-for-flow`, or `--output json` and locating the row by `flowId`) makes the subject the row under test rather than whatever page 1 holds.

e2e_tests/triggers_usecases_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+	require.Contains(t, stdout, triggersTestTriggerID, "the trigger id is missing from the listing")
```
jymaire: These two `Contains` assertions are independent, so the #118 regression can slip past them: `every_min` and `io.kestra.plugin.core.trigger.Schedule` may be satisfied by two **different** rows. Any other flow on the instance carrying a Schedule trigger — including a leftover from an earlier run of this very suite, since nothing is cleaned up — makes the type assertion pass even if the row under test rendered its TYPE column blank, which is precisely the symptom #118 was.

Asserting against the single row for this flow closes it: `--output json`, find the entry whose `flowId == flowID`, and check its `type` (and `id`) on that one object.

e2e_tests/triggers_usecases_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+	require.Contains(t, stdout, triggersTestTriggerID, "the trigger id is missing from the output")
+	require.Contains(t, stdout, triggersTestNamespace, "the namespace is missing from the output")
+}
+
+// TestTriggers_backfillOps_decode covers the other half of issue #131. Whether
+// a backfill exists to act on differs by server line — 1.x answers 404 for a
+// trigger that has none, 2.0 accepts the call — so the assertion is only that
+// the response is read rather than rejected over a missing property.
+func TestTriggers_backfillOps_decode(t *testing.T) {
```
jymaire, line 201: Unlike the other two tests, this one calls `RunAuthenticatedCliCmd` immediately after `flows deploy` with no wait for the trigger to be registered, and discards `err` — the only assertion is the absence of `no value given for required property`. If the scheduler has not yet picked up the freshly deployed flow, the server answers not-found and all three subtests go green without ever decoding a trigger body. The decode regression this test exists to catch could then re-land undetected, and the outcome varies run to run.

Waiting for the trigger to appear first (reusing `eventuallyTriggerListed`) before running the backfill ops makes the subject deterministic.

e2e_tests/triggers_usecases_test.go
AcevedoR: Fixed in f5ca6aa. The listing is now paged through with `--output json` (`--page N --size 200`, stopping on a short page) until the row whose `flowId`/`namespace` match the flow this test deployed, so the assertion no longer depends on page 1 holding it.

e2e_tests/triggers_usecases_test.go
AcevedoR: Fixed in f5ca6aa. Both assertions are now `require.Equal` against the single `triggerRow` for this flow, decoded from `--output json` — `triggerId` and `type` on the same object, so another flow's Schedule trigger can no longer satisfy the type check.

e2e_tests/triggers_usecases_test.go
AcevedoR: Fixed in f5ca6aa. The backfill test now waits for `triggers search-for-flow` to list the trigger (same poll as the other tests) before running the three ops, so the ops always hit a registered trigger and actually decode a body. `err` is still discarded on the ops themselves: a 1.x server legitimately answers 404 for a trigger with no backfill.

jymaire (approved):

PR #139, opened 2026-09-04
fix(logs): require --trigger-id on logs delete-flow

### LINKED ISSUE #133: logs delete-flow without --trigger-id is rejected by the server (triggerId required)
## Summary
`kestractl logs delete-flow <ns> <flow>` documents `--trigger-id` as optional, but the server rejects the request when it is omitted, on **both** Kestra 1.3.37 and 2.0.0-rc14:
```
$ kestractl logs delete-flow qa.ns qa-hello
Error: API error: Invalid query parameter: Required QueryValue [triggerId] not specified    # 2.0 wording
Error: API error: Bad Request: Required QueryValue [triggerId] not specified               # 1.3 wording
```
kestractl `main` (`6bb00bc`, go-sdk v2.0.0-rc4).

## Evidence
```
DELETE /api/v1/main/logs/qa.ns/qa-hello              -> 400
DELETE /api/v1/main/logs/qa.ns/qa-hello?triggerId=x  -> 200   (raw curl, 1.3.37)
```

## Options
- send the parameter always (empty value) if the endpoint accepts it, or
- mark `--trigger-id` required and say so in the help text, or
- implement "all logs of a flow" through the execution-scoped delete.

## Suggested e2e
`TestLogs_deleteFlow_withoutTrigger` asserting exit 0 and an empty `logs search --flow-id` afterwards.

Found by the `/kestractl-qa` matrix run of 2026-09-04 (T4.14).

AcevedoR:
## Summary

`kestractl logs delete-flow <ns> <flow>` documented `--trigger-id` as optional, but the server rejects the request without it. This makes the flag required, rejects a missing value locally before any request is sent, and corrects the help text to describe what the endpoint actually does.

## Evidence

Raw curl against two local Kestra EE instances — `localhost:9802` (**1.3.35**) and `localhost:9801` (**2.0.0-rc13**) — after deploying a one-task flow and running it to completion so the flow definitely had logs:

| request | 1.3.35 | 2.0.0-rc13 |
| --- | --- | --- |
| `DELETE /api/v1/main/logs/<ns>/<flow>` | `400 Bad Request: Required QueryValue [triggerId] not specified` | `400 Invalid query parameter: Required QueryValue [triggerId] not specified` |
| `DELETE …?triggerId=` (empty) | `200` | `200` |
| `DELETE …?triggerId=nope` | `200` | not retested (same handler) |

The decisive part is what the accepted calls actually did: **nothing.** Before and after `?triggerId=`, `GET /api/v1/main/logs/<executionId>` returned the identical single log entry on both versions:

```
--- logs before
[{"namespace":"e2e.dbg133","flowId":"dbg133","taskId":"hello", … "message":"hello from dbg133"}]
--- DELETE ?triggerId=   status=200
--- logs after
[{"namespace":"e2e.dbg133","flowId":"dbg133","taskId":"hello", … "message":"hello from dbg133"}]
```

I also checked whether a *real* trigger id deletes a flow's logs, using a flow with a `Schedule` trigger (`id: sched`, `cron: '* * * * *'`) on 1.3.35. After two scheduled executions, `logs/search` reported `"total":2`; after `DELETE …?triggerId=sched` returned `200`, it still reported `"total":2`. The logs of trigger-launched executions carry no `triggerId` field at all.

So `triggerId` is a genuine exact-match filter on the trigger a log entry is attributed to, not merely a required-but-ignored parameter — and execution logs are never attributed to a trigger. This endpoint cannot delete "all logs for a flow" under any argument.

## Decision

**Option (b): make `--trigger-id` required, with an accurate help text and a clear local error.**

Option (a) is ruled out by the evidence above: an empty value is accepted with a `200` on both versions but deletes nothing, so "always send the parameter" would turn a visible `400` into a silent no-op reported to the user as `deleted` — strictly worse than the bug.

Option (c) (implementing "all logs of a flow" over the execution-scoped delete) is a new feature — it would need to enumerate the flow's executions and delete each one's logs, with its own paging, partial-failure and confirmation semantics. That is out of scope for a bug fix and deserves its own issue; #133 is about the command not matching its own documentation.

Changes:

- `--trigger-id` is validated in `runLogsDeleteFlow` and in the command's `RunE` **before the client is built**, so no request is sent and the user gets:
  `--trigger-id is required: the server only deletes the logs attributed to a specific trigger of the flow. To delete an execution's logs, use 'kestractl logs delete <execution_id>'`
- `Short` / `Long` / `Example` and the flag description now say the command deletes a *trigger's* logs, and point at `logs delete <execution_id>` for execution logs.
- The signature drops the now-meaningless `*string`, and the success line plus JSON payload name the trigger (`triggerId` added to the JSON object — additive).

## Tests

Written test-first. With the signature change in place but no validation, the new tests failed for the intended reason:

```
--- FAIL: TestRunLogsDeleteFlow_MissingTriggerIDFailsWithoutRequest
    logs_test.go:222: expected an error when --trigger-id is omitted
--- FAIL: TestLogsDeleteFlowCommand_MissingTriggerID
    logs_test.go:269: expected a --trigger-id error, got: client error
```

New unit tests in `src/cli/logs_test.go` (httptest, following the `newTestClient` pattern):

- `TestRunLogsDeleteFlow_MissingTriggerIDFailsWithoutRequest` — asserts the error **and** that the handler saw zero requests.
- `TestRunLogsDeleteFlow_SendsTriggerID` — pins `DELETE /api/v1/main/logs/my.ns/my-flow?triggerId=my-trigger`.
- `TestLogsDeleteFlowCommand_MissingTriggerID` — the flag error wins over a failing client, i.e. validation really is pre-request.

`TestLogsDeleteFlowCommand_ClientError` now passes `--trigger-id`, since it can no longer reach the client without it.

New e2e test `TestLogs_deleteFlow_withoutTrigger` in a **new** file `e2e_tests/logs_delete_usecases_test.go` with its own namespace `e2e.logsdelete` and helper `deployLogsDeleteFlow`, to avoid conflicting with #136's `e2e_tests/logs_usecases_test.go`.

Commands run:

```
$ go build -o kestractl .                      # ok
$ go test ./src/...                            # ok  github.com/kestra-io/kestractl/src/cli
$ go vet ./src/...                             # clean
$ gofmt -l src/cli                             # src/cli/apps.go, src/cli/test_suites.go (pre-existing, untouched here)
$ cd e2e_tests && go vet ./... && gofmt -l .   # clean

$ cd e2e_tests && go test -run TestLogs_deleteFlow_withoutTrigger -v ./...   # against 2.0.0-rc13
--- PASS: TestLogs_deleteFlow_withoutTrigger (3.64s)
    --- PASS: TestLogs_deleteFlow_withoutTrigger/without_--trigger-id (0.52s)
    --- PASS: TestLogs_deleteFlow_withoutTrigger/with_--trigger-id (0.76s)
```

Both paths were also exercised by hand against **1.3.35** (the e2e suite pins `e2eHost` to `:9801`, so 1.3 could not be driven through `go test` locally):

```
$ ./kestractl logs delete-flow e2e.dbg133 dbg133 --host http://localhost:9802 …
Error: --trigger-id is required: the server only deletes the logs attributed to a specific
trigger of the flow. To delete an execution's logs, use 'kestractl logs delete <execution_id>'
exit=1

$ ./kestractl logs delete-flow e2e.dbg133 dbg133 --trigger-id e2e-trigger --host http://localhost:9802 …
Logs for trigger 'e2e-trigger' of flow 'dbg133' in namespace 'e2e.dbg133' deleted.
```

### Not verified

- The e2e job in CI is what will run `TestLogs_deleteFlow_withoutTrigger` across the full version matrix; locally it ran only against 2.0.0-rc13.
- That deleting by a trigger id ever removes *anything* — I could not produce a log entry carrying a `triggerId` with the plugin-less image (a `Schedule` trigger logs nothing of its own). The command's contract is unchanged for that case; the fix only concerns the missing parameter.
- `go.mod` / `go.sum` untouched.

Fixes #133

### DIFF
--- a/e2e_tests/logs_delete_usecases_test.go
+++ b/e2e_tests/logs_delete_usecases_test.go
@@ -1,0 +1,75 @@
+package e2e_tests
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// Coverage for 'logs delete-flow' (#133).
+//
+// The flow-scoped log deletion endpoint declares triggerId as a *required*
+// query parameter on both server lines, so the command used to fail with a raw
+// server 400 whenever --trigger-id was omitted — while its help text advertised
+// deleting "all logs for a flow". Verified by raw curl on 1.3.35 and
+// 2.0.0-rc13:
+//
+//	DELETE /api/v1/main/logs/<ns>/<flow>              -> 400 (triggerId not specified)
+//	DELETE /api/v1/main/logs/<ns>/<flow>?triggerId=   -> 200, but deletes nothing
+//
+// An empty value is therefore not a usable stand-in: it is a silent no-op. The
+// command now rejects a missing --trigger-id itself, before any request, and
+// this test pins that on every version in the matrix.
+
+const logsDeleteTestNamespace = "e2e.logsdelete"
+
+// deployLogsDeleteFlow deploys the one-task subject flow. --override keeps a
+// rerun against a persistent instance from being a hard failure, and
+// io.kestra.plugin.core.log.Log is available in the plugin-less image this
+// suite runs against.
+func deployLogsDeleteFlow(t *testing.T, flowID string) {
+	t.Helper()
+
+	source := fmt.Sprintf(`id: %s
+namespace: %s
+tasks:
+  - id: hello
+    type: io.kestra.plugin.core.log.Log
+    message: kestractl e2e logs delete
+`, flowID, logsDeleteTestNamespace)
+
+	path := filepath.Join(t.TempDir(), flowID+".yml")
+	require.NoError(t, os.WriteFile(path, []byte(source), 0o600))
+
+	stdout, stderr, err := RunAuthenticatedCliCmd(t, "flows", "deploy", path, "--override")
+	require.NoError(t, err, "deploy failed\nstdout: %s\nstderr: %s", stdout, stderr)
+	require.Contains(t, stdout, "1 flow(s) deployed successfully")
+}
+
+// TestLogs_deleteFlow_withoutTrigger asserts the command fails locally with its
+// own message when --trigger-id is omitted, and reaches the endpoint
+// successfully when it is supplied.
+func TestLogs_deleteFlow_withoutTrigger(t *testing.T) {
+	flowID := "e2e-logs-delete-flow"
+	deployLogsDeleteFlow(t, flowID)
+
+	t.Run("without --trigger-id", func(t *testing.T) {
+		stdout, stderr, err := RunAuthenticatedCliCmd(t, "logs", "delete-flow", logsDeleteTestNamespace, flowID)
+
+		require.Error(t, err, "expected a non-zero exit\nstdout: %s\nstderr: %s", stdout, stderr)
+		require.Contains(t, stderr, "--trigger-id is required")
+		// The point of the fix: kestractl answers, the server is never asked.
+		require.NotContains(t, stderr, "triggerId] not specified")
+	})
+
+	t.Run("with --trigger-id", func(t *testing.T) {
+		stdout, stderr, err := RunAuthenticatedCliCmd(t,
+			"logs", "delete-flow", logsDeleteTestNamespace, flowID, "--trigger-id", "e2e-trigger")
+
+		require.NoError(t, err, "stdout: %s\nstderr: %s", stdout, stderr)
+		require.Contains(t, stdout, "deleted")
+	})
+}
--- a/src/cli/logs.go
+++ b/src/cli/logs.go
@@ -1,6 +1,7 @@
 package cli
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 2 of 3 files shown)


AcevedoR (commented):
Reviewed as PM.

- The curl evidence settles it: `triggerId=` (empty) returns 200 on both 1.3 and 2.0 and deletes nothing, so sending it would turn a visible 400 into a false "deleted". Requiring the flag with a clear pre-request error is the honest option; the execution-scoped "delete all logs of a flow" is a separate feature.
- Help/Short/Long/Example now describe what the endpoint actually does. JSON payload gains `triggerId` (additive). The status line wording changed, which is fine for a command that never succeeded without the flag.
- Validation lives in `runLogsDeleteFlow` (testable) and is also short-circuited in RunE before a client is built.
- Unit + full e2e matrix green. New e2e file avoids overlapping #136.

Ready to merge from my side.

jymaire (approved):
## Approving

Verified on the PR head: `go vet ./src/...` and `go vet ./e2e_tests/...` clean, `go test ./src/...` passes (1144). CI is green on all legs, including the v2.0.0-rc13 and `develop` e2e matrix.

The change is correct. `--trigger-id` is validated before `NewClient()` so nothing is sent when it is missing, the `*string` → `string` signature change has no other callers, `triggerId` is now sent unconditionally with a non-empty value, and the new `triggerId` JSON field is additive. The e2e stderr assertion cannot false-positive: the repo does not set `SilenceErrors`, so Cobra writes `Error: …` to stderr, and the usage block it also dumps contains `(required)` but not the literal message asserted on. Making the flag required is a user-visible contract change, but it is called out in the description and justified by the raw-curl evidence on both server lines — an omitted `triggerId` is a 400 and an empty one is a silent no-op, so the previous signature had no usable behaviour to preserve.

Four non-blocking notes, none of which need to hold up the merge:

1. **`src/cli/logs.go:160` — the success line asserts more than the server confirms.** By this PR's own evidence the endpoint answers `200` for *any* `triggerId`, including one that matched nothing, so `logs delete-flow ns flow --trigger-id typoo` prints "Logs for trigger 'typoo' … deleted." and exits 0. Since the point of the PR is that this command was misrepresenting the endpoint's semantics, phrasing this as a request ("Deletion of logs for trigger … requested.") would close the last gap.

2. **`e2e_tests/logs_delete_usecases_test.go:70` — the happy-path subtest carries little signal.** It passes `e2e-trigger`, which the deployed flow does not define, and only asserts stdout contains "deleted", so it pins the message from (1) rather than any deletion behaviour and would keep passing against a no-op endpoint. `TestRunLogsDeleteFlow_SendsTriggerID` is the test doing the real work here.

3. **`src/cli/logs.go:125` — the empty check is duplicated in `RunE` and `runLogsDeleteFlow`,** and the rest of the repo expresses required flags with `cmd.MarkFlagRequired(...)` (10 call sites, e.g. `src/cli/executions.go:1245`). The custom message is a fair reason to hand-roll it, but the flag is then not marked required for shell completion and the two copies can drift; `MarkFlagRequired` alongside the explanation already in `Long` would cover both.

4. **`src/cli/logs.go:105` — no `-y/--yes` confirmation on a destructive command,** which AGENTS.md's review checklist asks for and every other destructive command in the repo has (`apps.go`, `flows.go`, `roles.go`, …). Pre-existing for both `logs delete-flow` and `logs delete`, but this PR is reworking the command's contract, so it is a natural place to add it or to scope it out explicitly.

PR #140, opened 2026-09-04
fix(executions): send `run --wait` progress to stderr

### LINKED ISSUE #135: executions run --wait writes progress lines to stdout, breaking -o json
## Summary
`kestractl executions run <ns> <flow> --wait -o json` prints two prose lines before the JSON document, so the output cannot be parsed. Reproduced on Kestra EE **2.0.0-rc14** and **1.3.37** (also rc13 / 1.3.35) with kestractl `main` (`6bb00bc`, go-sdk v2.0.0-rc4).

```
$ kestractl executions run qa.ns qa-hello --wait -o json | jq .id
Triggering execution of flow 'qa-hello' in namespace 'qa.ns'...
Waiting for execution to complete...
jq: parse error: Invalid numeric literal at line 1, column 11
```

Without `--wait` the JSON is clean. Table mode is affected as well: the two lines go to stdout, not stderr.

## Cause
`runExecutionsRun` (`src/cli/executions.go:971-975`) writes the progress messages with `fmt.Fprintf(renderer.Writer(), …)` before the API call. `renderer.Writer()` is the command's stdout, the same stream the JSON is rendered to afterwards, and the prints run regardless of the selected output format. These are the only two progress prints in the codebase that use the renderer's writer this way.

## Impact
`--wait` is the flag a CI script would use (trigger, block until terminal, read the state from JSON). Today the only scriptable path is to drop `--wait` and poll `executions get`. Note `--wait` also exits 0 when the execution ends `FAILED`, which compounds the problem for automation (separate concern, not filed).

## Fix
Send progress to stderr, matching the version-check notice that already uses `cmd.ErrOrStderr()` in `root.go`. `runExecutionsRun` only receives the `Renderer`, so either pass the error writer in or add an `ErrWriter()` to `Renderer` next to `Writer()`. Suppressing the lines in JSON mode would also work but hides useful feedback during a long server-side wait.

## Suggested tests
- e2e: `TestExecutionsRun_waitJSONIsParseable` in `e2e_tests/executions_usecases_test.go` — run `--wait -o json`, unmarshal stdout, assert `.id` non-empty and `.state.current == "SUCCESS"`.
- unit: `runExecutionsRun` with a JSON renderer over a buffer, assert the buffer is valid JSON.

Found by the `/kestractl-qa` matrix run of 2026-09-04 (T2.1b).

AcevedoR:
## Summary

`kestractl executions run <ns> <flow> --wait -o json` emitted two prose progress lines before the JSON document, so `| jq .id` failed with a parse error. Table mode had the same problem in milder form: the progress was on stdout, mixed into the result. The progress lines now go to stderr; stdout carries only the rendered execution.

```
$ kestractl executions run e2e.executions e2e-executions-run-json --wait -o json 2>/dev/null | jq -r .id
zmMkiqi6hbLQpSJdI3OzW
```

## Root cause

`runExecutionsRun` (`src/cli/executions.go`) wrote both messages with `fmt.Fprintf(renderer.Writer(), …)`. `renderer.Writer()` is the command's stdout — the very stream `renderer.Render` then writes the JSON (or table) to.

A grep confirms these were the only two progress prints using the renderer's writer this way. The other two `renderer.Writer()` callers — `flows get` printing flow source and `flows usage-report` printing markdown — write actual output, not progress, and are untouched.

## Fix

- `Renderer` gained an optional `errOut` writer next to `out`, exposed as `ErrWriter()` (defaulting to `os.Stderr`) and settable via a chainable `WithErrWriter()`. This is additive: `NewRenderer` / `NewRendererFromFlags` signatures are unchanged, so no other command or test had to move, and `runX` stays testable with two buffers.
- `newExecutionsRunCommand` wires `cmd.ErrOrStderr()` into the renderer, matching how the version-check notice in `root.go` already reaches stderr.
- The two progress lines print via `renderer.ErrWriter()`. They still print — they are useful during a long wait.

**The result lines did not move.** `Execution ID: …`, `Flow: …`, `Namespace: …` and `State: …` are still rendered to stdout inside `renderer.Render`, so the `require.Contains(t, stdout, "State: SUCCESS")` pattern used by `e2e_tests/executions_usecases_test.go` (and, per the issue, by the pending #136 / #139 log suites) keeps working unchanged.

**No e2e assertion needed adjusting.** The only e2e caller of `executions run --wait` is `runTerminatedExecution`, which reads the id from stdout by regex and asserts `State: SUCCESS` in stdout — it does not assert an empty stderr. No `require.Empty(t, stderr)` exists on any `executions run` path.

Exit-code behaviour on a FAILED execution is explicitly **out of scope** here, as the issue notes.

## Tests

Written first, confirmed red for the right reason (`stdout is not valid JSON: invalid character 'T' looking for beginning of value`, with the two progress lines shown ahead of the JSON), then green.

New unit tests in `src/cli/executions_test.go`:
- `TestRunExecutionsRun_WaitJSONOutputIsParseable` — `--wait` with a JSON renderer over a buffer; `json.Unmarshal` of stdout must succeed and carry the execution id.
- `TestRunExecutionsRun_WaitProgressGoesToErrWriter` — both progress lines land on the error buffer, neither reaches stdout, and `Execution ID:` / `SUCCESS` still do.

New e2e test `TestExecutionsRun_waitJSONIsParseable` in `e2e_tests/executions_usecases_test.go` — runs `--wait --output json` for real, unmarshals stdout, asserts `.id` non-empty and `.state.current == "SUCCESS"`.

Commands run:

| command | result |
| --- | --- |
| `go build -o kestractl .` | ok |
| `go test ./src/...` | ok |
| `gofmt -l src/cli` | only pre-existing `apps.go`, `test_suites.go` (untouched by this PR) |
| `go vet ./src/...` | clean |
| `cd e2e_tests && go vet ./...` + `gofmt -l .` | clean |
| `cd e2e_tests && go test -run TestExecutionsRun_waitJSONIsParseable` | PASS (4.8s) against local Docker EE 2.0.0-rc13 on :9801 |
| `cd e2e_tests && go test -run TestExecution` | ok (29.4s) — whole executions suite, no regression |

Manual verification of `--wait -o json \| jq` on both server lines: EE 2.0.0-rc13 (:9801) and 1.3.35 (:9802) — both return a clean id and `SUCCESS`. Table mode manually checked too: result on stdout, the two progress lines on stderr.

Not verified: the full e2e matrix (`sh run-e2e-tests.sh`) across every version in `COMPATIBLE_KESTRA_VERSION.properties`; only the two locally running instances were exercised. CI covers the rest.

Fixes #135

### DIFF
--- a/e2e_tests/executions_usecases_test.go
+++ b/e2e_tests/executions_usecases_test.go
@@ -1,6 +1,7 @@
 package e2e_tests
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -208,4 +209,27 @@ func TestExecutionsBulk_reportsTheAffectedCount(t *testing.T) {
 	require.NoError(t, err, "stdout: %s\nstderr: %s", stdout, stderr)
 	require.Contains(t, stdout, "1 execution(s) affected",
 		"the affected count was dropped at decode\nstdout: %s", stdout)
+}
+
+// TestExecutionsRun_waitJSONIsParseable pins #135: `--wait` used to print its
+// two progress lines to stdout, so `-o json` output could not be piped into a
+// JSON parser. The progress now goes to stderr and stdout stays machine
+// readable.
+func TestExecutionsRun_waitJSONIsParseable(t *testing.T) {
+	flowID := "e2e-executions-run-json"
+	deployExecutionsTestFlow(t, flowID)
+
+	stdout, stderr, err := RunAuthenticatedCliCmd(t, "executions", "run", executionsTestNamespace, flowID, "--wait", "--output", "json")
+	require.NoError(t, err, "run failed\nstdout: %s\nstderr: %s", stdout, stderr)
+
+	var execution struct {
+		ID    string `json:"id"`
+		State struct {
+			Current string `json:"current"`
+		} `json:"state"`
+	}
+	require.NoError(t, json.Unmarshal([]byte(stdout), &execution),
+		"stdout is not parseable JSON:\n%s", stdout)
+	require.NotEmpty(t, execution.ID)
+	require.Equal(t, "SUCCESS", execution.State.Current)
 }
--- a/src/cli/executions.go
+++ b/src/cli/executions.go
@@ -954,6 +954,7 @@ the execution completes (SUCCESS, FAILED, or other terminal state).`,
 			if err != nil {
 				return err
 			}
+			renderer.WithErrWriter(cmd.ErrOrStderr())
 			client, err := NewClient()
 			if err != nil {
 				return err
@@ -970,8 +971,10 @@ the execution completes (SUCCESS, FAILED, or other terminal state).`,
 
 func runExecutionsRun(client *Client, namespace, flowID string, wait bool, renderer *Renderer) error {
 	if wait {
-		fmt.Fprintf(renderer.Writer(), "Triggering execution of flow '%s' in namespace '%s'...\n", flowID, namespace)
-		fmt.Fprintln(renderer.Writer(), "Waiting for execution to complete...")
+		// Progress goes to stderr so it never mixes into the rendered output,
+		// which would make `--output json` unparseable.
+		fmt.Fprintf(renderer.ErrWriter(), "Triggering execution of flow '%s' in namespace '%s'...\n", flowID, namespace)
+		fmt.Fprintln(renderer.ErrWriter(), "Waiting for execution to complete...")
 	}
 
 	// POST /executions/{namespace}/{flowId} is shaped exactly like a 1.x action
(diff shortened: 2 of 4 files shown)


AcevedoR (commented):
Reviewed as PM.

- Minimal, additive fix: `Renderer` gains an optional error writer (defaults to stderr), constructor signatures unchanged, only the two progress lines move. The `State:` / `Execution ID:` result lines stay on stdout, so the `require.Contains(stdout, "State: SUCCESS")` pattern used by the suite and by open PRs #136/#139 is unaffected.
- Exit-code-on-FAILED behaviour deliberately left alone (separate concern per the issue).
- Unit tests were red first with the exact repro (`invalid character 'T'`), e2e parses `--wait -o json` on the whole matrix. All green.

Ready to merge from my side.

jymaire (approved):

PR #145, opened 2026-09-07
test: keep the test suites off the network

AcevedoR:
Tests should not depend on hosts we do not control. Two gaps here — one latent, one live.

## Unit suite: the convention was unenforced

`src/cli` already routes every external endpoint through a package var (`telemetryConfigURL`, `newVersionCheckLatestReleaseURL`, `pluginsAPIBase`, `pluginsMavenBase`) so tests can point it at an `httptest` server. Nothing enforced that. A test that forgot the override still passed while quietly calling `api.kestra.io`, `api.github.com` or `repo1.maven.org` — surfacing later as a flake on an offline machine or a rate-limited runner.

`src/cli/main_test.go` adds a `TestMain` that installs a dialer rejecting any non-loopback address, with an error naming the address and the vars to override. Two details worth knowing:

- The transport is **mutated in place** rather than replaced. `newCompatHTTPClient` captures `http.DefaultTransport` as its base at construction time, so swapping the package var would have left the SDK path unguarded.
- `Proxy` is nilled, otherwise an ambient `HTTP_PROXY` turns every external request into a loopback dial and slips past the guard.

`TestNetworkGuard_BlocksExternalDialsAndAllowsLoopback` keeps the guard from silently degrading into a no-op. It is hermetic: `http.Transport` hands the dialer the hostname, so the rejection happens before any DNS lookup.

**No existing test was blocked** — the suite was already clean. This only keeps it that way.

## E2E suite: a live leak

`isolatedEnv` strips every `KESTRACTL_*` variable to keep the developer's own credentials out of a run. That also stripped `KESTRACTL_TELEMETRY_DISABLED`, so every CLI invocation in the e2e suite fetched the PostHog config from `api.kestra.io` and shipped a `cli_command_executed` event to the **production** PostHog project.

Both switches are now set *after* the filter loop. The update notifier was already inert there (`go build` with no ldflags leaves `version == "dev"`, which `newVersionCheckSkipped` bails on), but it is set explicitly so it does not come back to life if the harness ever stamps a version.

## Verification

- `go test ./src/...` — pass, including `-race`
- `gofmt` / `go vet` clean on both modules (the pre-existing `gofmt` diffs in `apps.go` and `test_suites.go` are untouched and unrelated)
- Guard confirmed to block external and allow loopback via its own test

`test:` prefix, so no version bump.

### DIFF
--- a/src/cli/main_test.go
+++ b/src/cli/main_test.go
@@ -1,0 +1,103 @@
+package cli
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"strings"
+	"testing"
+	"time"
+)
+
+// TestMain fences the unit suite off from the network.
+//
+// Several commands talk to hosts we do not control — api.kestra.io for the
+// PostHog telemetry config, api.github.com for the update notifier,
+// repo1.maven.org for plugin downloads — and every test that exercises them is
+// expected to point the corresponding package var at an httptest server. That
+// convention is invisible until it is broken: a test that forgets an override
+// still passes, quietly making a real request, and only shows up later as a
+// flake on an offline machine or a rate-limited CI runner.
+//
+// So the dialer under the tests refuses anything that is not loopback. A missed
+// override fails with a message naming the address instead of succeeding by
+// accident.
+func TestMain(m *testing.M) {
+	blockExternalNetwork()
+	os.Exit(m.Run())
+}
+
+// errExternalNetwork is returned for any non-loopback dial attempt.
+type errExternalNetwork struct {
+	network string
+	address string
+}
+
+func (e *errExternalNetwork) Error() string {
+	return fmt.Sprintf(
+		"unit tests must not reach the network: blocked %s dial to %s. "+
+			"Point the relevant package var (telemetryConfigURL, newVersionCheckLatestReleaseURL, pluginsAPIBase, pluginsMavenBase, ...) at an httptest server",
+		e.network, e.address,
+	)
+}
+
+func blockExternalNetwork() {
+	transport, ok := http.DefaultTransport.(*http.Transport)
+	if !ok {
+		panic("http.DefaultTransport is not *http.Transport; cannot install the network guard")
+	}
+
+	// Mutated in place on purpose: newCompatHTTPClient captures
+	// http.DefaultTransport as its base at construction time, so replacing the
+	// package var would leave the SDK path unguarded.
+	dialer := &net.Dialer{Timeout: 5 * time.Second}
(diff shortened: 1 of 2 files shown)


jymaire (approved):

src/cli/main_test.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// TestNetworkGuard_BlocksExternalDialsAndAllowsLoopback keeps TestMain honest.
+// Without it the guard could become a silent no-op — a changed transport type, a
+// client that builds its own — and nothing else in the suite would notice.
+//
+// The external case is itself hermetic: http.Transport hands the dialer the
+// hostname, so the guard rejects it before any DNS lookup happens.
+func TestNetworkGuard_BlocksExternalDialsAndAllowsLoopback(t *testing.T) {
+	if _, err := http.Get("https://api.github.com/repos/kestra-io/kestractl/releases/latest"); err == nil {
```
jymaire: The guard's self-test issues a real request to `api.github.com`. That is safe only while the guard works — the moment it regresses (the exact failure this test exists to catch), the test makes a live request to api.github.com from CI, the very thing this PR forbids, and only *then* fails on the second assertion.

Pointing it at an unroutable target keeps the failure mode offline-safe while still exercising the dialer:

```go
if _, err := http.Get("http://192.0.2.1/repos/kestra-io/kestractl/releases/latest"); err == nil {
```

(`192.0.2.1` is TEST-NET-1; a `.invalid` host works too.)

src/cli/main_test.go
```diff
@@ -0,0 +1,103 @@
...
+// errExternalNetwork is returned for any non-loopback dial attempt.
+type errExternalNetwork struct {
+	network string
+	address string
+}
+
+func (e *errExternalNetwork) Error() string {
+	return fmt.Sprintf(
+		"unit tests must not reach the network: blocked %s dial to %s. "+
+			"Point the relevant package var (telemetryConfigURL, newVersionCheckLatestReleaseURL, pluginsAPIBase, pluginsMavenBase, ...) at an httptest server",
```
jymaire, line 99: The guard blocks the dial, but a missed override does not actually *fail a test* for two of the four package vars named here. `fetchPosthogConfig` (`src/cli/telemetry.go:127`) discards the client error and returns `"", ""`, and `newVersionCheck`'s fetch is equally best-effort — so a test that forgets to point `telemetryConfigURL` or `newVersionCheckLatestReleaseURL` at an httptest server still passes, silently, just without the outbound call. The enforcement holds only for paths that propagate the error (e.g. `plugins.go`'s `http.DefaultClient.Do`).

Either narrow the message to say so, or have the guard record blocked addresses and fail in `TestMain` after `m.Run()` if any were hit — that would cover the swallowed-error paths too.

PR #148, opened 2026-09-09
fix(ci): publish images to GHCR only until Docker Hub is authorized

AcevedoR:
The `v2.1.0` release failed. The Docker Hub **login succeeded**, but the push was refused:

```
failed to push kestra/kestractl:2.1.0: failed to authorize: failed to fetch oauth token:
401 Unauthorized: access token has insufficient scopes
```

The CI token is valid but not authorized to push to `kestra/kestractl`, and the repository does not exist for the push to auto-create — a repository-scoped token cannot create one. Fixing that is a Docker Hub account action, tracked separately; this PR unblocks releases in the meantime.

### Why it took the whole release down

`dockers_v2` publishes *before* the GitHub release pipe (`internal/pipe/publish/publish.go`), so the Docker failure aborted the run before the release was created:

| | |
|---|---|
| Tag `v2.1.0` | exists (`ef0c87b`) |
| GitHub release for `v2.1.0` | **missing** — newest release is still `2.0.0` |
| `ghcr.io/kestra-io/kestractl` | `2.1.0`, `2.1.0-static` ✅ |
| `docker.io/kestra/kestractl` | nothing |
| Rolling tags | not applied |

So `curl … | bash` still installs v2.0.0. GHCR worked fine — the images pushed successfully before the Docker Hub leg failed — so this drops Docker Hub and keeps everything else.

### Four places had to move together

A registry listed in one place but not the others either fails or silently tests nothing:

- **`.goreleaser.yml`** — `kestra/kestractl` commented out of both `dockers_v2` entries, with the re-enable steps in the comment.
- **`release.yml`** — the rolling-tag loop no longer iterates Docker Hub. It would have run `regctl image copy` against a source manifest that was never pushed and **failed the next release too**, for a different reason.
- **`tests.yml`** — the smoke test pulled `kestra/kestractl:<v>-amd64`, which snapshot builds no longer produce, so `docker-build` would have failed. The image name is now derived once rather than repeated four times.
- **`README.md` / `AGENTS.md`** — documented as GHCR-only, with what to undo.

### Also fixes review finding #2

The incident vindicated it, so it is folded in: the Docker Hub login is now skipped when the secret is absent instead of hard-failing. It runs before GoReleaser, so a missing or rotated secret would otherwise take down the binary release as well.

Note the suggested `if: ${{ secrets.DOCKERHUB_USERNAME != '' }}` cannot work — `secrets` is not available in a step-level `if` (nor a job-level one) and raises `Unrecognized named-value`. The secrets are mapped into workflow `env` and the condition tests `env` instead.

### Verification

- `goreleaser check` clean; snapshot builds all four images, **all** under `ghcr.io`, none under `kestra/kestractl`.
- The smoke-test commands run green against the built images (version, `--help`, `git`/`curl`/`bash`, static variant).
- Rolling-tag script emits GHCR targets only — for `v2.1.1` with `DOCKER_LATEST=true`: `2.1`, `2`, `latest` plus the three `-static` forms.
- `render-compat-badges.sh --check` and `go test ./src/...` pass.

### After merge

This is a `fix:`, so Auto Tag will cut **`v2.1.1`** and release it. Retrying `v2.1.0` is not an option: `workflow_dispatch` checks out the tag's tree, which still has Docker Hub enabled. The `v2.1.0` tag stays orphaned with no release behind it — delete it or leave it, your call.

Worth a follow-up: the failed run burned **28 minutes** because GoReleaser retried the push 10 times with backoff on an auth error that could never succeed. Capping `dockers_v2`'s `retry` would fail fast.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -45,6 +45,10 @@ env:
   # Unlike GitHub's pointer this one is NOT inert on a prerelease, so the rolling
   # tag step below keeps an rc off `:latest` itself.
   DOCKER_LATEST: "true"
+  # Mapped here so the login step can be conditioned on them; `secrets` is not
+  # available in a step `if`.
+  DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
+  DOCKERHUB_PASSWORD: ${{ secrets.DOCKERHUB_PASSWORD || secrets.DOCKERHUB_TOKEN }}
 
 jobs:
   release:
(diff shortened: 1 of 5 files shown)


AcevedoR:
Superseded: the Docker Hub token has been granted push access to `kestra/kestractl`, so disabling the registry is no longer needed. Recovering `v2.1.0` instead by re-running Release with `tag=v2.1.0` — the tag's tree still has Docker Hub enabled, which is now what we want.

The review finding folded in here (skip the Docker Hub login when the secret is absent, since `secrets` is not available in a step-level `if`) is still worth having and will come back in a follow-up alongside the other open findings.

PR #154, opened 2026-09-09
fix(plugins): verify downloads with the strongest published checksum

AcevedoR:
Follow-up to #149. The scanners are live and reporting; these are the two open findings, plus a correction to what I claimed about the Go toolchain.

## First, a correction

In #149 I said the shipped binary carried **8 reachable stdlib CVEs** and that `go-version: "1.25"` needed bumping to 1.26. **That was wrong**, and no toolchain change is needed. Two mistakes on my side: I read my local toolchain (1.26.3, itself out of date) as representative, and I sorted the Go release list lexicographically, so `go1.25.9` looked like the newest 1.25 when the real latest is **1.25.14**.

Verified properly this time, by building with each toolchain and scanning the binary with `govulncheck -mode=binary`:

| binary built with | result |
|---|---|
| `go1.25.9` (my arbitrary local pick) | 10 reachable stdlib vulns, fixed in 1.25.10 / 1.25.13 |
| `go1.25.14` (what `go-version: "1.25"` actually resolves to) | **No vulnerabilities found** |

CI already builds on a patched stdlib, and the fixes were available within the 1.25 line all along. The follow-up PR I proposed isn't needed, and I've left the pins alone.

## What the scanners actually found

`govulncheck` 0, CodeQL 0, Trivy 0 on both images. The two open alerts are from **Plerion**, which was already installed on the repo:

### `fix(plugins)` — SHA-1 only checksum verification (CWE-328, medium)

Plugin JARs were verified against the `.sha1` file alone. SHA-1 is not collision resistant, so a substituted artifact can be crafted to match a digest that a merely corrupted download would fail.

Now negotiated strongest-first: `.sha512` → `.sha256` → `.sha1`, verifying against the first one the repository actually publishes. SHA-1 is kept as a last resort rather than dropped — Maven Central has only published the strong sums since 2020, and private repositories (what EE installs are usually pointed at) frequently carry `.sha1` alone, where refusing it would mean *no* verification at all.

One non-obvious part, and the reason the diff isn't just a hash swap: a checksum body that isn't a hex digest of the expected width is treated as "not published" and falls through to the next algorithm. Repositories and caching proxies answer a missing checksum file with 200 and the *artifact itself*, or an HTML error page — the repo's own test mocks did exactly this — and trusting that as a digest would fail every download.

Scope worth being explicit about: this does **not** defend against a hostile repository, since the checksum comes from the same host as the JAR. It defends against a corrupted or substituted artifact.

Cost: usually unchanged at one checksum request per artifact (Central publishes `.sha512`); three against a repository that only has `.sha1`.

### `fix(bindings)` — no input validation on bulk create (CWE-502, medium)

The scanner's stated rationale does not apply to Go: `encoding/json` into `interface{}` yields only maps, slices and scalars, with no type instantiation or gadget chains. **This was not an unsafe-deserialization vulnerability.**

The missing validation it pointed at was real, though. `--file` is documented as an array of binding requests, but the bare `interface{}` target forwarded any JSON at all to the API — a single object, a bare string, a number — and the complaint came back as an opaque server error. It was also the only bare `interface{}` unmarshal target left in `src/cli`; every other site already uses `map[string]any`.

Now `[]map[string]any` with an empty-array check. Deliberately *not* the SDK's generated request model: a slice of maps forwards fields verbatim, so a server accepting a key this SDK build doesn't model keeps working, and the model's enum/required-field validation can't reject a request the server would have accepted — the kind of client-side strictness that has bitten this repo before.

**Behaviour call-out:** input that isn't a JSON array is now rejected locally instead of being sent. Such a request would have failed server-side anyway, so this trades an opaque 400 for a clear local error, but it is a stricter input contract on `bindings bulk-create`.

## Verification

- `go build`, `go vet`, `gofmt` clean; **1174 tests pass**, plus `-race` on the plugin tests.
- 6 new tests: SHA-512 preferred (and `.sha1` never requested), the full fallback ladder when strong sums are 404, fallback when a proxy returns 200 with a non-digest body, SHA-512 mismatch rejected with cleanup, per-algorithm digest correctness, and the four rejected `--file` shapes.
- The `checksum mismatch` error prefix is preserved, with the algorithm added — `checksum mismatch (SHA-512) for …`.
- Test mocks now answer checksum requests through one shared helper that behaves like a real repository, instead of nine hand-rolled `.sha1` branches.

Both commits are `fix(...)`, so merging this cuts a patch release — intended for a security fix under the #149 rules.

## Not fixed

Nothing outstanding from the scanners. `Vulnerabilities Checks` was also dispatched manually to exercise the Trivy image jobs that only run on schedule — both images clean.

### DIFF
--- a/src/cli/bindings.go
+++ b/src/cli/bindings.go
@@ -363,9 +363,18 @@ func runBindingsBulkCreate(client *Client, path string, renderer *Renderer) erro
 		return fmt.Errorf("failed to read file: %w", err)
 	}
 
-	var requests interface{}
+	// A slice of maps rather than the SDK's generated request model: the fields
+	// are forwarded to the API verbatim, so a server that accepts a key this
+	// SDK build does not model still works. It is also not a bare interface{} —
+	// that would accept any JSON at all (an object, a bare string, a number)
+	// and defer the complaint to an opaque server-side error, when --file is
+	// documented as an array of binding requests.
+	var requests []map[string]any
 	if err := json.Unmarshal(data, &requests); err != nil {
-		return fmt.Errorf("failed to parse JSON file: %w", err)
+		return fmt.Errorf("failed to parse JSON file: %w (expected a JSON array of binding requests)", err)
+	}
+	if len(requests) == 0 {
+		return fmt.Errorf("no binding requests found in %s: expected a non-empty JSON array", path)
 	}
 
 	bindings, err := client.Kestra.Bindings().BulkCreateBinding(client.Ctx, client.Tenant, requests)
--- a/src/cli/bindings_test.go
+++ b/src/cli/bindings_test.go
@@ -343,3 +343,33 @@ func TestRunBindingsBulkCreate(t *testing.T) {
 		t.Errorf("expected binding ID in output, got:\n%s", buf.String())
 	}
 }
+
+func TestRunBindingsBulkCreate_RejectsNonArrayFile(t *testing.T) {
+	cases := map[string]string{
+		"a single object": `{"type":"USER","externalId":"u1","roleId":"r1"}`,
+		"a bare string":   `"USER"`,
+		"an empty array":  `[]`,
+		"malformed JSON":  `[{"type":`,
+	}
+	for name, body := range cases {
+		t.Run(name, func(t *testing.T) {
+			f, err := os.CreateTemp("", "bindings-*.json")
+			if err != nil {
+				t.Fatal(err)
+			}
+			defer os.Remove(f.Name())
+			f.WriteString(body)
+			f.Close()
+
+			// No server: a shape this wrong must fail before any request.
+			var buf bytes.Buffer
+			err = runBindingsBulkCreate(newTestClient(t, "http://127.0.0.1:1"), f.Name(), newTableRenderer(&buf))
+			if err == nil {
+				t.Fatalf("expected an error for %s, got nil", name)
+			}
+			if !strings.Contains(err.Error(), "JSON array") {
+				t.Errorf("expected the error to explain the expected shape, got: %v", err)
+			}
+		})
+	}
+}
--- a/src/cli/plugins.go
+++ b/src/cli/plugins.go
@@ -3,10 +3,13 @@ package cli
 import (
 	"context"
 	"crypto/sha1"
+	"crypto/sha256"
+	"crypto/sha512"
 	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
+	"hash"
 	"io"
 	"net/http"
 	"os"
@@ -27,6 +30,33 @@ var errRateLimited = errors.New("rate limited by repository (429)")
 // rateLimitWaits defines the back-off delays between successive 429 retries.
 // Exposed as a var so tests can override it to avoid sleeping.
 var rateLimitWaits = []time.Duration{5 * time.Second, 10 * time.Second, 30 * time.Second}
+
+// checksumAlgorithm describes one of the checksum files a Maven repository
+// publishes alongside an artifact.
+type checksumAlgorithm struct {
+	ext    string // checksum file extension, without the dot
+	name   string // human-readable name, for log messages
+	hexLen int    // length of the hex-encoded digest
+	new    func() hash.Hash
+}
+
+// checksumAlgorithms lists the checksum files we accept, strongest first.
+//
+// SHA-1 is not collision resistant, so it is the last resort rather than the
+// only option: a repository that publishes .sha512 or .sha256 gets verified
+// with that instead. It cannot simply be dropped — Maven Central has only
+// published the stronger sums since 2020, and private repositories (including
+// the ones an EE install is usually pointed at) frequently carry .sha1 alone,
+// where refusing it would mean no verification at all.
+//
+// Note this is defence against a corrupted or substituted artifact, not against
+// a hostile repository: the checksum comes from the same host as the JAR, so an
+// attacker who controls the response controls both.
+var checksumAlgorithms = []checksumAlgorithm{
+	{ext: "sha512", name: "SHA-512", hexLen: 128, new: sha512.New},
+	{ext: "sha256", name: "SHA-256", hexLen: 64, new: sha256.New},
+	{ext: "sha1", name: "SHA-1", hexLen: 40, new: sha1.New},
+}
 
 // pluginsAPIBase and pluginsMavenBase are vars so tests can point them at a local httptest server.
 var (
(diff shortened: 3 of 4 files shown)


AcevedoR:
## QA against the live registry — this PR had a regression, now fixed

Pushed 0d0a341. The question was what happens to an existing user whose `plugins/` directory is already full of JARs. Answering it properly turned up a real problem with the ladder as I first wrote it.

### Existing plugin directories: untouched

`downloadJAR` returns at the `os.Stat` cache check **before** any checksum work, so an already-present JAR is never re-fetched or re-verified. Confirmed live against `repo1.maven.org`, old binary populating the directory and the new one running over it:

```
old, fresh dir:      io.kestra.plugin:plugin-kafka:1.10.2 ... done (76 MB)
new, same dir:       io.kestra.plugin:plugin-kafka:1.10.2 ... already up to date
```

Zero network calls for checksums, zero risk of an existing directory failing after upgrade. Fresh and `--force-redownload` downloads produce byte-identical JARs to the old binary.

**Separate pre-existing gap worth knowing:** because the cache check is `os.Stat` only, a *corrupted* existing JAR is also reported `already up to date`. I truncated one to 10 bytes and it passed. That is true of the old binary too — no behaviour change here — but it means existing JARs are never verified, and `--force-redownload` is the only way to re-verify a directory. Not fixed in this PR: re-hashing every JAR on every run is a real cost and a behaviour change that deserves its own decision.

### The regression: Central publishes no strong sums for Kestra artifacts

My PR description claimed Central has published the stronger sums since 2020. **That is wrong for exactly the artifacts kestractl downloads.** Measured:

| artifact | .sha512 | .sha256 | .sha1 | .md5 |
|---|---|---|---|---|
| `io.kestra.plugin:plugin-kafka:1.10.2` | 404 | 404 | 200 | 200 |
| `io.kestra.plugin:plugin-aws:2.8.5` | 404 | 404 | 200 | 200 |
| `io.kestra.plugin:plugin-jdbc-postgres:2.0.2` | 404 | 404 | 200 | 200 |
| `io.kestra.plugin:plugin-git:4.0.0` | 404 | 404 | 200 | 200 |
| `io.kestra:core:2.0.0` | 404 | 404 | 200 | 200 |
| `org.apache.kafka:kafka-clients:3.9.0` | 200 | 200 | 200 | 200 |
| `com.squareup.okhttp3:okhttp:4.12.0` | 200 | 200 | 200 | 200 |

Some publishers do, Kestra does not. So the ladder as first written probed two 404s per artifact and then verified with **the same `.sha1` as before** — no added strength at all, double the requests. The 1.3.9 compatibility set is **214 plugins** and `--concurrency` defaults to **1**, so a plain `plugins download` paid **428 extra requests and ~31s** (72ms measured per 404 to Central) against a registry that rate limits — which is why `downloadJAR` carries 429 back-off in the first place.

### The fix: negotiate once per repository

Absences are now cached per repository base URL for the lifetime of one command. Measured over 20 plugins against a mock mirroring Central's real responses:

| binary | total | .sha512 | .sha256 | .sha1 | jars |
|---|---|---|---|---|---|
| before this series | 40 | 0 | 0 | 20 | 20 |
| this PR, as first pushed | 80 | 20 | 20 | 20 | 20 |
| **now** | **42** | **1** | **1** | 20 | 20 |

Extrapolated to the real 214-plugin set: 428 → 856 → **430**.

And the case the ladder exists for still works fully — against a repository that publishes `.sha512`, all 10 artifacts verify with it and `.sha1` is never consulted, at no extra request cost. That is the case EE customers on an internal Nexus/Artifactory are in.

**Trade-off, documented at the type:** on a host with mixed availability, one artifact's 404 makes the rest of the run fall back to `.sha1`. Measured and shown in the tests. That equals the behaviour before this series rather than degrading it, and a single run downloads one publisher's artifacts.

### Honest summary of what this PR now buys

For the default Maven Central path with Kestra plugins: **no change in verification strength** — still SHA-1, because that is all Central publishes — at a cost of 2 extra requests per command. The security improvement is real only against repositories that publish strong sums. Worth keeping for that and for the non-digest-body hardening, but I don't want the PR title overselling it.

### Verification

- `go build`, `go vet`, `gofmt` clean; full suite passes; `-race -count=2` on the plugin tests.
- 4 new tests: negotiation happens once per repository not per artifact, strong sums are still used for every artifact where published, the shared negotiator is concurrency-safe under `--concurrency 8`, and the cache is keyed per repository *and* extension (nil-safe).
- Live: existing-directory skip, fresh download, and `--force-redownload` all byte-identical to the old binary against `repo1.maven.org`.

AcevedoR:
Closing this. The premise doesn't hold up, and the QA above is most of the reason.

**The finding isn't a security issue.** The `.sha1` file and the `.jar` are fetched from the same host over the same TLS connection, so anyone who can alter one can alter the other. The checksum is a corruption check, not a security control, and upgrading its hash function only matters to an attacker who can substitute the JAR but *not* the checksum file — an asymmetry that doesn't occur in practice. Exploiting SHA-1 here would need a chosen-prefix collision *plus* that asymmetry. TLS already covers transport integrity.

I had been holding onto the change "for internal Nexus/Artifactory targets, which do publish strong sums". That argument is wrong for the same reason: a digest served by a Nexus is just as same-origin as one served by Central. There is no repository against which the ladder buys security.

Two corrections to what I claimed earlier in this PR:

- **"Maven Central has published the stronger sums since 2020"** — not for the artifacts kestractl downloads. Measured: `.sha512` and `.sha256` are 404 for `plugin-kafka`, `plugin-aws`, `plugin-jdbc-postgres`, `plugin-git` and `io.kestra:core`; only `.sha1` and `.md5` exist. Other publishers (`kafka-clients`, `okhttp`) do serve them, Kestra does not.
- **"Non-digest body hardening"** cited as independent value — it wasn't. The pre-existing code already validated digest length and hex and skipped verification otherwise. That check only became necessary because the fallthrough I introduced needed it.

So what was left was ~200 lines, a negotiator with a documented mixed-host fallback, 10 tests and +2 requests per command, in exchange for nothing on any real path. Not worth carrying.

**What is a real gap:** Kestra plugins are GPG-signed on Central — `plugin-kafka-1.10.2.jar.asc` returns 200 with a valid PGP signature — and kestractl ignores it. A signature is bound to the publisher's key rather than the host, so it defends against exactly what a checksum cannot: a compromised registry, a hostile mirror, a MITM with a valid certificate. Filed as #156.

The `fix(bindings)` commit from this branch is independently useful and continues as #155. The Plerion CWE-328 alert is dismissed as won't-fix with the same-origin reasoning recorded on it; it can be reopened from the Security tab if anyone disagrees.

Download-path behaviour on `main` is therefore unchanged from before this PR.

PR #162, opened 2026-09-10
feat(flows): add validate-by-query to validate the flows stored on an instance

### LINKED ISSUE #161: flows validate: add --remote to validate the flows already stored on an instance
### Feature request

Add a way to validate the flows **already stored on an instance**, rather than only local files.

`kestractl flows validate <path>` today takes a local file or directory. There is no command that answers "are all the flows on this instance valid?" — which is the exact question users have after upgrading to 2.0 and migrating their flow definitions.

### Why

We want to document a simple post-migration verification step in the official docs. With what exists today, the shortest correct recipe is a three-step round-trip through the filesystem:

```bash
kestractl flows export-by-query --output-file flows.zip
unzip -q -o flows.zip -d flows/
kestractl flows validate ./flows/
```

That works (and `validate` already exits non-zero on constraint violations, so it gates CI), but requiring `unzip` and a scratch directory to answer a read-only question is a poor first impression for a migration-day check.

### Proposal

Add a flag to the existing command so the sources are fetched from the server instead of the filesystem:

```bash
kestractl flows validate --remote                          # every flow on the instance
kestractl flows validate --remote --namespace company.team # scoped
```

Implementation is a composition of two calls that already exist in the SDK client:

1. `FlowsAPI.ExportFlowsByQuery` (same call `flows export-by-query` makes, honouring the existing `--namespace` / `--flow` / `--filter` by-query flags)
2. unzip in memory, then feed the sources into the existing `runFlowsValidate` path → `FlowsAPI.ValidateFlows`

Reuses the existing renderer and exit-code behaviour, so `--output json` and CI gating come for free.

`--remote` and a positional `<path>` should be mutually exclusive.

Suggested detail: `export/by-query` deliberately skips drafts (a draft-headed flow exports at its last saved revision, and a draft-only flow is not exported at all). Worth stating in the command help so `--remote` is not read as "validates literally everything".

### Note for large instances

`runFlowsValidate` joins all sources into a single request body. That is already true for a big local directory, but `--remote` makes it easy to hit accidentally on an instance with thousands of flows. Consider chunking the validation into batches rather than one request — or, at minimum, documenting `--namespace` as the workaround.

### Related gap

There is a second post-migration failure mode this does **not** cover: a flow whose stored YAML the server cannot deserialize at all (e.g. a task type removed in 2.0). The server returns those from `/flows/search` as a `FlowWithException`, carrying an extra `exception` field. `kestractl flows search` cannot surface them, because `runFlowsSearch` projects the response down to `id`, `namespace`, `revision` and `description` — so even `--output json` drops `exception`. Today that check has to be done with raw `curl` + `jq`.

Happy to raise that as its own issue if it should be tracked separately.

AcevedoR:
Closes #161.

Answering "is what is on the instance valid?" needed a three-step round-trip through the filesystem:

```bash
kestractl flows export-by-query --output-file flows.zip
unzip -q -o flows.zip -d flows/
kestractl flows validate ./flows/
```

This adds `flows validate-by-query`, which does it in one command, in memory:

```bash
kestractl flows validate-by-query --namespace company.team  # scoped
kestractl flows validate-by-query --all                     # every flow on the instance
kestractl flows validate-by-query --all --output json       # CI gate, exits non-zero
```

Named after the existing `-by-query` family, so it takes the same `--namespace` / `--flow` / `--filter` flags as `export-by-query`. A selection is required — with no flags the command errors and prints the help, since validating a whole instance should be asked for by name. `flows validate <path>` is unchanged.

It finds the same flows the three-step recipe does — the win is ergonomics, not extra detection.

Drafts are not covered: the export skips them, so a draft-headed flow is validated at its last saved revision.

## Changes

- **new** `src/cli/flows_validate_by_query.go` + tests.
- `flows_usage_analysis.go` — `flowSourcesFromZip` extracted to keep archive entry names; `flowsFromZip` is now a wrapper, `usage-report` untouched.
- `flows.go` — split `formatValidateResults` into a builder and a renderer, and added `validateSourcesInBatches` (100 sources per request, violation indices offset per batch). `flows validate` uses it too, so large local directories stop going out as one request.
- `README.md` — the command, plus a "Verifying a migration" section.

No breaking changes: `flows validate`'s table output is byte-identical and `ValidateResult`'s JSON shape is unchanged.

## Testing

1186 unit tests, `go vet`, `gofmt`, clean under `-race`.

Live QA against EE 2.0.0 on an instance migrated in place from 1.3.37 — [full report and matrix in the comment below](https://github.com/kestra-io/kestractl/pull/162#issuecomment-5620663814). It caught two bugs, both fixed here: export archives are flat (`namespace-flow.yml`, not `namespace/flow.yaml`), and broken flows were reported twice. Tenant handling (`--tenant`, `KESTRACTL_TENANT`, flag-over-env) verified across two tenants.

### DIFF
--- a/src/cli/flows.go
+++ b/src/cli/flows.go
@@ -559,25 +559,15 @@ func listAllFlows(client *Client) ([]parsedFlow, error) {
 // listAllFlowsForTenant pages through every flow of the given tenant. Commands
 // that scan more than the configured tenant call it directly.
 func listAllFlowsForTenant(client *Client, tenant string) ([]parsedFlow, error) {
-	return listAllFlowsForTenantFiltered(client, tenant, nil)
-}
-
-// listAllFlowsForTenantFiltered is listAllFlowsForTenant narrowed to the flows
-// matching the shared *-by-query filters. Nil filters means the whole tenant,
-// which is what the endpoint does by default.
-func listAllFlowsForTenantFiltered(client *Client, tenant string, filters []kestra.QueryFilter) ([]parsedFlow, error) {
 	const pageSize int32 = 1000
 	page := int32(1)
 	results := make([]parsedFlow, 0)
 
 	for {
-		req := client.API.FlowsAPI.SearchFlows(client.Ctx, tenant).
+		resp, _, err := client.API.FlowsAPI.SearchFlows(client.Ctx, tenant).
 			Page(page).
-			Size(pageSize)
-		if len(filters) > 0 {
-			req = req.Filters(filters)
-		}
-		resp, _, err := req.Execute()
+			Size(pageSize).
+			Execute()
 
 		var batch []parsedFlow
 		var total int64
@@ -615,11 +605,6 @@ type parsedFlow struct {
 	Description string
 	Revision    int32
 	Source      string
-	// Exception is set when the server could not deserialize the stored flow
-	// source and answered with a FlowWithException. Such a flow exists in the
-	// inventory but is absent from an export, so it is invisible to any check
-	// that only reads exported sources.
-	Exception string
 }
 
 // parsedFlowFromFlow normalizes a decoded SDK Flow into a parsedFlow.
@@ -629,24 +614,7 @@ func parsedFlowFromFlow(f kestra.Flow) parsedFlow {
 		Namespace:   f.GetNamespace(),
 		Description: f.GetDescription(),
 		Revision:    f.GetRevision(),
-		// "exception" is not in the generated Flow model; the SDK keeps such
-		// fields in AdditionalProperties.
-		Exception: flowException(f.AdditionalProperties),
 	}
-}
-
-// flowException reads the "exception" field the server adds to a flow it could
-// not deserialize. It is a string in practice, but the field is untyped, so
-// anything else is rendered rather than dropped.
-func flowException(m map[string]any) string {
-	value, ok := m["exception"]
-	if !ok || value == nil {
-		return ""
-	}
-	if s, ok := value.(string); ok {
-		return strings.TrimSpace(s)
-	}
-	return strings.TrimSpace(fmt.Sprint(value))
 }
 
 // parsedFlowFromMap extracts flow fields from a raw JSON object. Labels are
@@ -669,7 +637,6 @@ func parsedFlowFromMap(m map[string]any) parsedFlow {
 	if v, ok := m["revision"].(float64); ok {
 		f.Revision = int32(v)
 	}
-	f.Exception = flowException(m)
 	return f
 }
 
--- a/src/cli/flows.go
+++ b/src/cli/flows.go
@@ -1089,8 +1089,14 @@ func validateSourcesInBatches(client *Client, sources []string, batchSize int) (
 
 	for start := 0; start < len(sources); start += batchSize {
 		end := min(start+batchSize, len(sources))
+		count := end - start
 
-		body := strings.Join(sources[start:end], "\n---\n")
+		documents := make([]string, 0, count)
+		for _, source := range sources[start:end] {
+			documents = append(documents, trimDocumentMarkers(source))
+		}
+
+		body := strings.Join(documents, "\n---\n")
 		batch, _, err := client.API.FlowsAPI.ValidateFlows(client.Ctx, client.Tenant).
 			Body(body).
 			Execute()
@@ -1099,12 +1105,56 @@ func validateSourcesInBatches(client *Client, sources []string, batchSize int) (
 		}
 
 		for _, violation := range batch {
-			violation.SetIndex(violation.GetIndex() + int32(start))
+			index := violation.GetIndex()
+			// The server numbered a document this batch did not send, which
+			// means the body split into more documents than it has sources.
+			// Every index in the batch is then suspect, so report nothing
+			// rather than attach results to the wrong flows.
+			if index < 0 || int(index) >= count {
+				return nil, fmt.Errorf(
+					"cannot map validation results back to flows: the server reported document %d of a %d-document request, "+
+						"which happens when a flow source contains its own YAML document separator; "+
+						"re-run with --batch-size 1 to validate one flow per request",
+					index, count)
+			}
+			violation.SetIndex(index + int32(start))
 			violations = append(violations, violation)
 		}
 	}
 
 	return violations, nil
+}
+
+// trimDocumentMarkers strips the YAML document markers around a single flow
+// source so that joining sources with "\n---\n" yields exactly one document
+// per source.
+//
+// A stored flow source keeps whatever the author wrote, and a leading "---" is
+// ordinary YAML style. Joined as-is it produces "---\n---", i.e. an empty
+// document, which shifts every later index in the batch by one and makes a
+// valid flow fail with "No content to map due to end-of-input".
+func trimDocumentMarkers(source string) string {
+	lines := strings.Split(source, "\n")
+
+	start := 0
+	for start < len(lines) && strings.TrimSpace(lines[start]) == "" {
+		start++
+	}
+	if start < len(lines) && strings.TrimRight(lines[start], " \t") == "---" {
+		start++
+	}
+
+	end := len(lines)
+	for end > start {
+		last := strings.TrimRight(lines[end-1], " \t")
+		if last == "" || last == "---" || last == "..." {
+			end--
+			continue
+		}
+		break
+	}
+
+	return strings.Join(lines[start:end], "\n")
 }
 
 // formatValidateResults renders the violations of a local `flows validate`
@@ -1128,9 +1178,6 @@ func formatValidateResults(violations []kestra.ValidateConstraintViolation, file
 func buildValidateResults(violations []kestra.ValidateConstraintViolation, seeds []ValidateResult) ([]ValidateResult, int) {
 	results := make([]ValidateResult, len(seeds))
 	copy(results, seeds)
-	for i := range results {
-		results[i].Success = true
-	}
 
 	unknownResults := make([]ValidateResult, 0)
 
--- a/src/cli/flows.go
+++ b/src/cli/flows.go
@@ -71,6 +71,7 @@ func newFlowsCommand() *cobra.Command {
 	cmd.AddCommand(newFlowsEnableByQueryCommand())
 	cmd.AddCommand(newFlowsExportByIdsCommand())
 	cmd.AddCommand(newFlowsExportByQueryCommand())
+	cmd.AddCommand(newFlowsValidateByQueryCommand())
 	cmd.AddCommand(newFlowsGenerateGraphFromSourceCommand())
 	cmd.AddCommand(newFlowsGenerateGraphCommand())
 	cmd.AddCommand(newFlowsTaskCommand())
@@ -558,15 +559,25 @@ func listAllFlows(client *Client) ([]parsedFlow, error) {
 // listAllFlowsForTenant pages through every flow of the given tenant. Commands
 // that scan more than the configured tenant call it directly.
 func listAllFlowsForTenant(client *Client, tenant string) ([]parsedFlow, error) {
+	return listAllFlowsForTenantFiltered(client, tenant, nil)
+}
+
+// listAllFlowsForTenantFiltered is listAllFlowsForTenant narrowed to the flows
+// matching the shared *-by-query filters. Nil filters means the whole tenant,
+// which is what the endpoint does by default.
+func listAllFlowsForTenantFiltered(client *Client, tenant string, filters []kestra.QueryFilter) ([]parsedFlow, error) {
 	const pageSize int32 = 1000
 	page := int32(1)
 	results := make([]parsedFlow, 0)
 
 	for {
-		resp, _, err := client.API.FlowsAPI.SearchFlows(client.Ctx, tenant).
+		req := client.API.FlowsAPI.SearchFlows(client.Ctx, tenant).
 			Page(page).
-			Size(pageSize).
-			Execute()
+			Size(pageSize)
+		if len(filters) > 0 {
+			req = req.Filters(filters)
+		}
+		resp, _, err := req.Execute()
 
 		var batch []parsedFlow
 		var total int64
@@ -604,6 +615,11 @@ type parsedFlow struct {
 	Description string
 	Revision    int32
 	Source      string
+	// Exception is set when the server could not deserialize the stored flow
+	// source and answered with a FlowWithException. Such a flow exists in the
+	// inventory but is absent from an export, so it is invisible to any check
+	// that only reads exported sources.
+	Exception string
 }
 
 // parsedFlowFromFlow normalizes a decoded SDK Flow into a parsedFlow.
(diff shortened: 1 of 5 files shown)


AcevedoR:
## Live QA — Kestra EE 2.0.0 (GA), real in-place 1.3.37 → 2.0.0 migration

Run against `registry.kestra.io/docker/kestra-ee:v2.0.0` on the ee-sandbox, using the scenario this command is actually for: deploy 1.x-era flows on **EE 1.3.37** with a persisted store, then restart the same data directory on **EE 2.0.0** (`kestra.migration.auto=true`) so the flows are genuinely broken-in-place rather than synthetically broken. The server rejects invalid flows on write, so this is the only way to get one stored.

Corpus: `still-valid`, plus four 1.x constructs — `EachSequential`, `ForEach`, `pluginDefaults`, trigger `conditions`.

### Result

```
FLOW                                  STATUS  CONSTRAINTS
company.team/uses-each-sequential     FAILED  Validation error: Invalid type: io.kestra.plugin.core.flow.EachSequential
company.team/uses-foreach             FAILED  Validation error: Invalid type: io.kestra.plugin.core.flow.ForEach
company.team/uses-plugin-defaults     FAILED  Validation error: Unrecognized field "pluginDefaults" ...
company.team/uses-trigger-conditions  FAILED  Validation error: Unrecognized field "conditions" (class ...Schedule) ...
company.team/still-valid              OK      -

1 valid flow(s), 4 failed        exit 1
```

### Two real bugs found, both fixed in 6cb8926

**1. Archive layout was wrong.** A flow export is **flat** — entries are `company.team-valid-a.yml`, not `company.team/valid-a.yaml`. Since both the namespace and the flow id may contain a hyphen, that name cannot be split back apart. Identity is now read from the source YAML instead, and rows are labelled `<namespace>/<flow-id>`, which reads better than the archive filename and matches the cross-check rows.

**2. Broken flows were counted twice** — the migrated instance reported **6 failures for 4 broken flows**. The premise in the issue and in my own PR description was wrong: an undeserializable flow is *not* absent from the export. The export writes the stored source text back out without deserializing it, so the validate endpoint sees it and reports the same breakage in its own words, and the inventory cross-check then reported it a second time. The cross-check now skips flows the export already covered, so it only reports what the export genuinely missed. Command help and README corrected.

Regression test added for the double-count, with the live scenario in the comment.

### Matrix

| # | Check | Result |
|---|---|---|
| 1 | Tenant-wide, all valid (4 flows, clean 2.0 instance) | exit 0 |
| 2 | Migrated instance, 4 broken + 1 valid | exit 1, one row per flow |
| 3 | **Parity with the three-step recipe** (`export-by-query` → `unzip` → `validate ./dir/`) | identical: 1 valid, 4 failed |
| 4 | `--batch-size 1` (5 requests) | identical per-flow results |
| 5 | `--batch-size 2` (3 requests) | identical per-flow results |
| 6 | `--namespace company.team` | scoped correctly |
| 7 | `--flow still-valid` | 1 row, exit 0 |
| 8 | `--filter NAMESPACE:EQUALS:company.team` | scoped correctly |
| 9 | Namespace with no flows | `no flows matched the query` on stderr, exit 0 |
| 10 | Positional path rejected | `unknown command "./flows/"` |
| 11 | `--output yaml` | rejected by `validateOutputFormat()` |
| 12 | Regression: `flows validate ./dir/` | unchanged, 1 valid / 4 failed, exit 1 |
| 13 | Regression: `flows usage-report` (shares the refactored `flowsFromZip`) | clean report |

Batch sizes 1, 2 and 100 producing identical per-flow results is the real-server confirmation that the violation index offsetting is right — that was the part most likely to be silently wrong.

### Honest caveat

Item 3 is worth reading carefully: **on this instance the old three-step recipe found exactly the same 4 flows.** The cross-check added nothing here — it is a safety net for a flow the export cannot produce, which I could not reproduce naturally and which is covered only by a unit test with a synthetic `exception`. The real value of this PR over the recipe is ergonomic (no `unzip`, no scratch directory, scoping flags, one command), not extra coverage in the common case. I'd rather state that than let the issue's premise stand unchallenged.

Also not covered live: the permission-degradation path (inventory unreadable → warning on stderr, exported sources still validated). Unit test only.

Unit suite: `go vet`, `gofmt`, **1183 tests pass**, and the new tests pass under `-race`.

AcevedoR:
### Follow-up: inventory cross-check removed (84bdd37)

Acting on the caveat in the QA report above. The cross-check detected nothing the export path did not already report, so it is gone rather than kept as a speculative guard.

Re-verified on the same migrated 1.3.37 → 2.0.0 instance — detection is unchanged:

```
FAILED company.team/uses-each-sequential
FAILED company.team/uses-foreach
FAILED company.team/uses-plugin-defaults
FAILED company.team/uses-trigger-conditions
OK     company.team/still-valid

1 valid flow(s), 4 failed        exit 1
```

`--batch-size` 1 / 2 / 100 all still report 4 failures, and `--flow still-valid` still exits 0.

Removing it also reverted the plumbing that existed only to serve it — `parsedFlow.Exception`, the `flowException` helper, and the filtered variant of `listAllFlowsForTenant`. **`flows.go` now contains only the validate refactor**, which makes the diff a good deal easier to review. The command makes one export call plus one validate call per batch and never touches `/flows/search`; a test asserts that.

Net: 1183 tests, clean under `-race`.

src/cli/flows.go
```diff
@@ -1059,25 +1060,76 @@ func runFlowsValidate(client *Client, path string, recursive bool, renderer *Ren
...
+	if batchSize < 1 {
+		batchSize = validateBatchSize
+	}
+
+	violations := make([]kestra.ValidateConstraintViolation, 0, len(sources))
+
+	for start := 0; start < len(sources); start += batchSize {
+		end := min(start+batchSize, len(sources))
+
+		body := strings.Join(sources[start:end], "\n---\n")
```
jymaire: **Positional `\n---\n` join + index mapping can silently desync.** Joining sources with `"\n---\n"` and mapping violations back by `violation.GetIndex()` assumes each source is exactly one YAML document with no column-0 `---` line and no empty document. validate-by-query feeds *arbitrary/broken stored sources* to this endpoint, so if one flow's source contains a top-level `---` separator (or is empty), the server parses more documents than there are entries in `sources`, and every subsequent violation in the batch attaches to the wrong flow (or spills into synthetic `<index N>` rows). Nothing compares the server's document count to `len(sources)`, so the corruption is silent.

src/cli/flows_validate_by_query.go
```diff
@@ -0,0 +1,172 @@
...
+			skipped, maxZipEntrySize>>20)
+	}
+
+	sources := make([]string, 0, len(entries))
+	seeds := make([]ValidateResult, 0, len(entries))
+	for _, entry := range entries {
+		namespace, flowID := flowIdentityFromSource(entry.Source)
+		sources = append(sources, entry.Source)
+		seeds = append(seeds, ValidateResult{
+			FilePath:  flowLabel(namespace, flowID, entry.Name),
```
jymaire, line 121: **FLOW column is never updated from the server's returned identity.** `seeds[i].FilePath` is fixed here from the source-parsed identity via `flowLabel(...)`. `buildValidateResults` sets `target.Namespace`/`target.FlowID` from the server violation but never touches `FilePath`, and `renderValidateResults` prints `FilePath`. So for a flow whose top-level source can't be parsed, `flowLabel` falls back to the raw archive entry name (e.g. `company.team-legacy.yml`) and the FLOW column keeps showing that even though the server returned the real namespace/flow. The comment on `flowIdentityFromSource` ("the validate endpoint ... overwrites what is set here") is inaccurate: only the unused struct fields are overwritten, not the rendered label.

src/cli/flows.go
```diff
@@ -1059,25 +1060,76 @@ func runFlowsValidate(client *Client, path string, recursive bool, renderer *Ren
...
+const validateBatchSize = 100
+
+// validateSourcesInBatches validates sources in fixed-size batches and returns
+// the violations of all of them, re-indexed against the full source list.
+//
+// A violation identifies its source by position within the request body, so
+// every batch after the first comes back numbered from zero; the offset below
+// is what keeps a violation attached to the flow it actually belongs to.
+func validateSourcesInBatches(client *Client, sources []string, batchSize int) ([]kestra.ValidateConstraintViolation, error) {
+	if batchSize < 1 {
```
jymaire, line 1084: **`--batch-size 0`/negative is silently coerced to 100.** A user passing `--batch-size 0` (e.g. expecting "send everything in one request") gets 100 with no error or warning, so the actual behavior differs from what was asked with no feedback. Consider an explicit error (as `resolve()` does for an empty selection) or documenting the coercion.

src/cli/flows.go
```diff
@@ -1059,25 +1060,76 @@ func runFlowsValidate(client *Client, path string, recursive bool, renderer *Ren
...
+// results. A violation carries the positional index of the source it belongs
+// to within the validated batch, so seeds must be in the order the sources
+// were sent; an index outside that range lands in a synthetic row rather than
+// being dropped. A result fails only on a constraint — warnings, infos,
+// deprecations and the outdated flag are reported but never fail validation.
+func buildValidateResults(violations []kestra.ValidateConstraintViolation, seeds []ValidateResult) ([]ValidateResult, int) {
+	results := make([]ValidateResult, len(seeds))
+	copy(results, seeds)
+	for i := range results {
+		results[i].Success = true
```
jymaire: **Redundant re-initialization (dead code).** `results` was just `copy()`-ed from `seeds`, which always carry `Success: true`, and the trailing failed-count loop recomputes `Success` from `Constraints` for every row anyway. This `results[i].Success = true` loop has no effect and reads as if it guards something it does not.

src/cli/flows_usage_analysis.go
```diff
@@ -1449,16 +1449,40 @@ func (m *markdownWriter) printf(format string, args ...any) {
 // rather than buffered.
 const maxZipEntrySize = 10 << 20 // 10 MiB
 
+// zipFlowSource is one flow source read out of an export archive, together
+// with its archive entry name. Kestra names the entries
+// "<namespace>/<flow-id>.yaml", which is the only identity an OK flow has
+// before the server has said anything about it.
+type zipFlowSource struct {
```
jymaire, line 1457: **Contradictory comments on the archive entry naming format.** This comment says export entries are named `"<namespace>/<flow-id>.yaml"` (nested), while `flowIdentityFromSource`/`flowLabel` in `flows_validate_by_query.go` state Kestra *flattens* the archive to `"<namespace>-<flow-id>.yml"` and cannot split it back apart. One is wrong. The `flowLabel` fallback (used exactly when a source can't be parsed) and the skipped-entry warning depend on the true format, and the tests use both spellings, masking which the real server emits.

src/cli/flows.go
AcevedoR: Confirmed on a live EE 2.0.0 instance, and worse than silent mis-attribution — it is a false CI failure on a healthy instance.

Kestra stores the source verbatim, so a flow written with a leading `---` (ordinary YAML style) keeps it. Two **valid** flows, the separator one second in the archive:

```
company.sep2/aaa-first   OK
company.sep2/zzz-second  FAILED  Illegal Flow source: YAML parsing error:
                                 No content to map due to end-of-input
<index 2>                OK

2 valid flow(s), 1 failed        exit 1
```

Also reachable from the shipped `flows validate <path>`, which shares the join — the same two files as local YAML failed identically.

Fixed in 587ea3f: sources are trimmed of their surrounding document markers before the join, and an out-of-range index is now an error rather than a phantom row — one unmappable answer makes every index in the batch suspect, so guessing is worse than refusing. The interior-separator case you mention cannot be normalised away without changing meaning, and that guard is what catches it.

Same two flows now report `2 valid flow(s), 0 failed`, exit 0, at batch sizes 1, 2 and 100. Unit tests for `trimDocumentMarkers` (including `---` inside a string and indented in a block scalar, which must not be stripped) and for the out-of-range refusal.

src/cli/flows_validate_by_query.go
AcevedoR: Correct, including that the comment was inaccurate — the server overwrote two fields nothing rendered. Fixed in 587ea3f: `relabelFromServerIdentity` rewrites the label from the returned identity after the mapping, and the comment now says what actually happens. Test covers a source whose `id` cannot be parsed locally but that the server identifies.

src/cli/flows.go
AcevedoR: Agreed. Rejected at the flag level now (`--batch-size must be at least 1, got 0`), consistent with how the missing-selection error behaves, help included. The internal `batchSize < 1` guard stays as a defence for the in-process callers. 587ea3f.

src/cli/flows.go
AcevedoR: Confirmed dead — both call sites seed `Success: true` and the failed-count loop recomputes it. Removed in 587ea3f.

src/cli/flows_usage_analysis.go
AcevedoR: My error, and thanks for catching the reason it survived. The flat `<namespace>-<flow-id>.yml` is right — live QA established it and I fixed `flows_validate_by_query.go` at the time but missed this comment. Corrected in 587ea3f, and the zip fixtures now use the flat naming, since the nested spelling in the tests is exactly what let the stale comment through.

AcevedoR:
### Review fixes (587ea3f) — all five were correct

Thanks @jymaire, the first one was a genuine blocker.

**The `---` desync is a false CI failure, not just a mis-attribution.** Kestra stores a flow source verbatim, so a flow written with a leading `---` keeps it. Two **valid** flows, separator one second in the archive:

```
company.sep2/aaa-first   OK
company.sep2/zzz-second  FAILED  YAML parsing error: No content to map due to end-of-input
<index 2>                OK

2 valid flow(s), 1 failed        exit 1
```

**And it was already reachable from the shipped `flows validate <path>`**, which shares the join — the same two files as local YAML fail identically on `ed0e8a8`. So this PR now also fixes a pre-existing bug in a released command.

Fix: trim the document markers around each source before joining, and treat an index outside the batch as an error rather than rendering a phantom row — one unmappable answer makes every index in that batch suspect, so guessing is worse than refusing. An interior separator can't be normalised away without changing meaning; that guard is what catches it.

The other four are fixed as described in the thread replies: server identity now wins for the FLOW label, `--batch-size 0` is rejected, the dead `Success = true` loop is gone, and the stale nested-archive comment is corrected — along with the zip fixtures that used the nested spelling, which is what let it survive review in the first place.

**Verification.** Same two flows now `2 valid flow(s), 0 failed`, exit 0, at batch sizes 1, 2 and 100. Migrated 1.3.37 → 2.0.0 instance still finds its 4 genuinely broken flows out of 5, at every batch size. 1201 unit tests, `go vet`, `gofmt`, clean under `-race`. New tests: `trimDocumentMarkers` (including `---` inside a string and indented in a block scalar, which must *not* be stripped), the out-of-range refusal, server-identity relabelling, and the batch-size rejection.

jymaire:
**Optional / non-blocking** — nice command, works well. One design question:

`validate-by-query` exits **0 when zero flows match**. `runFlowsValidateByQuery` prints `no flows matched the query` to stderr, but with an empty result set `failed == 0`, so `renderValidateResults` returns `nil` and the process exits 0.

Since the README frames `--all --output json` as a CI migration gate, a mistyped scope (e.g. `--namespace compnay.team`) or an unexpectedly empty tenant would pass the gate green having validated nothing. If that's a deliberate "nothing invalid = pass" choice, all good — just flagging in case you'd prefer an empty match to exit non-zero (or warn louder) so a typo can't silently pass CI.

jymaire (approved):

PR #163, opened 2026-09-11
feat: manage reusable inputs blocks

brian-mulier-p:
Adds `kestractl reusable-inputs` (alias `ri`), covering every operation the
EE Reusable Inputs API exposes:

| Command | Endpoint |
|---|---|
| `list <namespace>` | `GET namespaces/{ns}/reusable-inputs` (`--page`, `--size`) |
| `get <namespace> <id>` | `GET namespaces/{ns}/reusable-inputs/{id}` (`--revision`) |
| `revisions <namespace> <id>` | `GET namespaces/{ns}/reusable-inputs/{id}/revisions` |
| `create <namespace> <id> --file` | `PUT …/{id}?failIfExists=true` |
| `update <namespace> <id> --file` | `PUT …/{id}` |
| `delete <namespace> <id>` | `DELETE …/{id}` |
| `namespaces` | `GET reusable-inputs/namespaces` |

`create` conflicts on an existing block, `update` saves a new revision. Both
PUT the file bytes verbatim: the server stores the request body as the block's
`source`, so a YAML round trip would reorder keys and break the stored source.

`get` prints the YAML source in table mode (like `flows get`) and the full
model — revision, `last`, timestamps — under `--output json`.

`list` and `get` resolve namespace inheritance, so a block defined in a parent
namespace is visible from its children; the NAMESPACE column and the JSON
output show which namespace owns each block. `revisions` and `delete` do not
resolve it — the help text for each says so.

Bumps `github.com/kestra-io/client-sdk/go-sdk/v2` to v2.0.1, the first release
carrying `ReusableInputsAPI`. The routes require a Kestra EE 2.0.1+ server.


AcevedoR (approved):
LGTM

AcevedoR (approved):

PR #165, opened 2026-09-11
test(flows): cover bulk-update delete default

### LINKED ISSUE #164: `flows bulk-update` defaults `--delete` to true, unlike `namespace-sync` (--delete opt-in) — inconsistent and dangerous default
Reported on [reddit](https://www.reddit.com/r/selfhosted/comments/1waruq6/comment/p94mcxq/?%24deep_link=true&correlation_id=a07a6a6a-40b7-4b26-86ef-e53bfafe97a7&ref=email_post_reply&ref_campaign=email_post_reply&ref_source=email&target_user=tchiotludo&%243p=e_as&_branch_match_id=1625893910911035599&utm_medium=Email+Amazon+SES&_branch_referrer=H4sIAAAAAAAAA3VO206EMBD9GvYNWAHBNSHGaPyNydAO0Gxpu9M26Ivf7hDdRzOd5PRccmZNKcTnumbS2qQKQ6iscde6DS9F07VhJMB4EujZLMahhcx2XI9U0b4WzYe8fd%2Brv7zymxAsG8nOq4%2BJtHyE3silKPBhR863XtCVYmKE5gwmgs8JCKMhhuThHgZ0GqJCS%2BIPl25Tn7ejs5XaThMFOI4t2vfEWSy98sxkMRnvwGjh8TxgL1N252kou6npy6ee5pIe22nGmS4DDpJjmsVMGxoL4ehlCvbrVwCFW0CzuP8d0WdWdNeFTMgLJciR%2BDhOrcYnm7U%2FfYudmI1bYGK%2Fiz6%2Brew3%2BgFaTYxIiAEAAA%3D%3D)
 
## Summary

`flows bulk-update` and `flows namespace-sync` do effectively the same thing (create/update every flow in a YAML file, optionally purging what's not in the file), but they disagree on the default for deletion:

```go
// flows.go, newFlowsBulkUpdateCommand
cmd.Flags().BoolVar(&deleteUnlisted, "delete", true, "Delete flows not present in the file")
```

```go
// flows.go, newFlowsNamespaceSyncCommand
cmd.Flags().BoolVar(&deleteMissing, "delete", false, "Delete flows in the namespace that are not in the file")
```

For `bulk-update`, deletion is opt-out. For `namespace-sync`, it's opt-in. Same shape of command, opposite defaults.

## Why this is a problem

`bulk-update` is the command most likely to be used in scripted/CI contexts (deploy a multi-doc YAML file). A truncated file, a bad merge, a template that only renders a subset of flows for a given environment, or a partial `cat` in a pipeline, will silently delete every flow omitted from the file, because deletion is the default rather than something the caller opted into.

There's precedent in the core `kestra` repo for treating this exact pattern as a footgun: kestra-io/kestra#1462 flipped the equivalent server-side namespace-update delete behavior from default-true to default-false, specifically because deleting flows by default was judged too dangerous. `bulk-update`'s current default reintroduces the same problem this CLI already avoids elsewhere (`namespace-sync`, and the Git `SyncFlows`/`SyncNamespaceFiles` plugin tasks, which also default `delete: false`).

## Suggested fix

Flip `bulk-update`'s default to `false`, matching `namespace-sync` and every other destructive-by-flag command in the CLI:

```go
cmd.Flags().BoolVar(&deleteUnlisted, "delete", false, "Delete flows not present in the file")
```

This is a breaking change for anyone relying on the implicit deletion, so it probably wants a changelog callout rather than a silent patch.

## Where

`src/cli/flows.go`, `newFlowsBulkUpdateCommand` — the flag definition, and the `Long` description above it, which currently documents the dangerous behavior as intentional ("will be deleted if --delete is set (default true)").

IzzaldinSamir:
## Summary

- Add direct regression coverage for the `flows bulk-update --delete` default.
- Assert that deletion remains opt-in (`false`) so a future flag change cannot silently restore destructive behavior.
- The production change was merged in #167; this PR is intentionally narrowed to the missing regression test.

## Testing

- `gofmt -d src/cli/flows_test.go`
- `go test -v ./src/cli/... -run '^TestFlowsBulkUpdate'`
- `go build ./...`
- `git diff --check upstream/main...HEAD`

`go test ./src/...` was also attempted on Windows. It reaches unrelated existing failures in the version-check tests that assume Unix home-path and `0600` permission behavior; all `TestFlowsBulkUpdate*` tests pass.

Related to #164 and #167.

### DIFF
--- a/src/cli/flows.go
+++ b/src/cli/flows.go
@@ -1731,8 +1731,8 @@ func newFlowsBulkUpdateCommand() *cobra.Command {
 		Use:   "bulk-update",
 		Short: "Update multiple flows from a YAML file (flows separated by ---).",
 		Long: `Create or update all flows defined in a multi-document YAML file.
-Flows already existing in the namespace but not present in the file will be deleted
-if --delete is set (default true).`,
+Flows already existing in the namespace but not present in the file can optionally
+be deleted with --delete.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if err := validateOutputFormat(); err != nil {
 				return err
@@ -1758,7 +1758,7 @@ if --delete is set (default true).`,
 
 	cmd.Flags().StringVarP(&filePath, "file", "f", "", "Path to YAML file with flows separated by --- (required)")
 	cmd.Flags().StringVarP(&namespace, "namespace", "n", "", "Target namespace for all flows")
-	cmd.Flags().BoolVar(&deleteUnlisted, "delete", true, "Delete flows not present in the file")
+	cmd.Flags().BoolVar(&deleteUnlisted, "delete", false, "Delete flows not present in the file")
 	cmd.Flags().BoolVar(&allowNamespaceChild, "allow-namespace-child", false, "Allow updating flows in child namespaces")
 
 	return cmd
--- a/src/cli/flows_test.go
+++ b/src/cli/flows_test.go
@@ -858,6 +858,16 @@ func TestFlowsBulkUpdateCommand_MissingFile(t *testing.T) {
 	}
 }
 
+func TestFlowsBulkUpdateCommand_DeleteIsOptIn(t *testing.T) {
+	deleteFlag := newFlowsBulkUpdateCommand().Flags().Lookup("delete")
+	if deleteFlag == nil {
+		t.Fatal("expected --delete flag to exist")
+	}
+	if deleteFlag.DefValue != "false" {
+		t.Errorf("expected --delete to default to false, got %q", deleteFlag.DefValue)
+	}
+}
+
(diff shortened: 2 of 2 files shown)


IzzaldinSamir:
Since #167 landed the behavior change, I rebased this PR onto current `main` and narrowed it to the remaining unique value: a regression test asserting that `flows bulk-update --delete` stays opt-in. The full `TestFlowsBulkUpdate*` group and `go build ./...` pass.

fdelbrayelle (approved):

fdelbrayelle:
Thank you @IzzaldinSamir :wave:

IzzaldinSamir:
Thank you for the review and merge! 👋

PR #171, opened 2026-09-16
docs: describe container tag shapes rather than v2 literals

AcevedoR:
Split out of #170, which found this while fixing the same staleness in the Docker Hub overview text.

## What

The README's container section names `2` / `2.0` / `2.0.0` in the tag table and pulls `:2` in every example. Those tags still resolve, but they stopped being the current line at `v3.0.0` and nothing updated the prose.

Rather than swapping `2` → `3`, the table now describes the tag *shapes* (`latest`, `X`, `X.Y`, `X.Y.Z`), with `:3` and `:3.1` named once as today's examples. It was stale one release after it was written; shapes do not need an edit at each major. `docker run` examples use `:latest`; the CI snippet keeps a literal major, since pinning is the point there.

This matches the wording in #170's `docs/docker-hub.md`, so the README and the Docker Hub overview say the same thing.

## Deliberately not touched

The installer section above still says the convenience script installs **"the latest kestractl v2 release"**. That is accurate — `install-scripts/install.sh:10` pins `DEFAULT_MAJOR="2"`, so a default `curl … | bash` really does resolve to the newest 2.x release, not 3.1.1.

So after this PR the README tells you to `docker pull` a 3.x image a few lines below telling you the installer gives you 2.x. **That inconsistency is real and this PR does not fix it** — it is `install.sh`'s default to change, not the prose, and bumping the installer's default major is a behavior change for every new user that deserves its own review rather than riding along in a docs PR.

Worth deciding separately: is `DEFAULT_MAJOR="2"` intentional (v2 held as the stable recommendation while 3.x settles) or simply missed when the major rolled? AGENTS.md documents the `VERSION=2` default in several places, all of which are currently accurate and would need updating alongside it.

## Checks

`render-compat-badges.sh --check` passes — the badge block is untouched.


fdelbrayelle:
superseded by my latest pr

PR #177, opened 2026-09-18
fix(secrets): use filtered tenant search to list namespace secrets

### LINKED ISSUE #172: secrets list: returns 403 Access Denied for every namespace
## Observation

`kestractl secrets list <namespace>` (`ls` alias), run with `-v` against a Kestra EE v2.0.2 server:

```
> GET http://localhost:8080/api/v1/default/namespaces/system/secrets?page=1&size=50
> Accept: application/json
> Authorization: REDACTED

< 403 Forbidden
< Content-Type: application/problem+json
< {"type":"https://kestra.io/docs/api-reference/problems/forbidden","title":"Access denied","status":403,"detail":"Access denied","instance":"/api/v1/default/namespaces/system/secrets"}
```

Reproduced for two namespaces (a reserved one and a normal one) and on both `kestractl` v2.2.0 and the current v3.1.1 release, using a token that succeeds on other commands against the same server (`namespaces list`, `flows list`, `kv list`, `users list` all return `200`).

The same server, token, and namespace filter succeed through a different route:

```
curl --globoff -H "Authorization: Bearer $KESTRA_TOKEN" \
  'http://localhost:8080/api/v1/default/secrets?filters[namespace][EQUALS]=system'
```

→ `200 OK`, `{"readOnly":false,"results":[],"total":0}`.

## Expected Behaviour

`kestractl secrets list <namespace>` should return the secrets for that namespace, as shown by the successful response to the equivalent filtered request above.

fdelbrayelle:
## Summary
- `secrets list <namespace>` called the SDK's `ListNamespaceSecrets` (`GET /namespaces/{namespace}/secrets`), which Kestra EE can 403 on even for a token that succeeds on every other namespace-scoped list (`namespaces list`, `flows list`, `kv list`, `users list`).
- The equivalent tenant-scoped, filtered search (`GET /secrets?filters[namespace][EQUALS]=<namespace>`) works with the same token. The generated Go SDK has no client for that endpoint, so `secrets list` now builds and decodes the request by hand via `Client.doRawRequest`, extended here to support a query string.

Fixes #172

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./src/...`
- [x] Added unit tests asserting `secrets list` hits `/secrets` with `filters[namespace][EQUALS]` (not `/namespaces/{namespace}/secrets`), plus empty-response and 403 cases.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


fdelbrayelle:
## QA report

Ran a live Kestra EE instance via `plugin-devtools/bin/kestra-core-run --ee --docker` (image `kestra-ee:develop`, v2.1.0-SNAPSHOT). Hit the known superadmin/tenant gotcha from AGENTS.md (`tenant-admin-access` missing → no tenant bound, `tenants list` returned 0 rows); recreated the container with `kestra.security.super-admin.tenant-admin-access: [main]` added, which auto-created and bound the `main` tenant.

**Reproduced the bug** on `main` against this instance:
```
> GET /api/v1/main/namespaces/system/secrets?page=1&size=50
< 403 Forbidden — {"title":"Access denied", ...}
Error: API error: Access denied
```

**Verified the fix** on this branch, same instance/token:
```
> GET /api/v1/main/secrets?filters%5Bnamespace%5D%5BEQUALS%5D=system&page=1&size=50
< 200 OK — {"readOnly":false,"results":[],"total":0}
```

Also exercised the full round trip on the fix branch:
- `secrets set system MY_TEST_KEY ...` → `secrets list system` (table + `-o json`) shows it correctly
- `secrets patch system MY_TEST_KEY --description ...` → list reflects the update
- `secrets delete system MY_TEST_KEY` → list goes back to empty
- Created a second, non-reserved namespace (`qa.namespace`) with its own secret and confirmed `filters[namespace][EQUALS]` isolates it — `secrets list system` stays empty while `secrets list qa.namespace` shows only its own secret
- `--page`/`--size` flags pass through correctly

`go build ./...`, `go vet ./...`, and `go test ./src/...` all pass.

No regressions found. LGTM.

src/cli/secrets.go
```diff
@@ -58,14 +69,19 @@ func runSecretsList(client *Client, namespace string, page, size int32, renderer
...
 	}
 
-	resp, _, err := client.API.NamespacesAPI.
-		ListNamespaceSecrets(client.Ctx, namespace, client.Tenant).
-		Page(page).
-		Size(size).
-		Filters([]kestra.QueryFilter{}).
-		Execute()
+	query := url.Values{}
+	query.Set("filters[namespace][EQUALS]", namespace)
```
jymaire, line 73: The namespace scoping now depends entirely on the server honoring filters[namespace][EQUALS] on the tenant-wide /secrets search, which the mocked tests cannot verify; if the secrets controller ignores that filter key, `secrets list <ns>` silently lists every namespace's secret metadata in the tenant.

The generated SDK serializes filters as a single CSV `filters` query param (parameterAddToHeaderOrQuery(..., "filters", ..., "form", "csv")), not as PHP-style nested keys, so there is no in-repo precedent proving the secrets controller accepts filters[namespace][EQUALS]. If Kestra EE's secrets endpoint doesn't apply that key, the request returns all namespaces' secrets and `secrets list child.ns` shows unrelated namespaces' keys/descriptions. AGENTS.md requires a live EE instance for QA; this path was only exercised against an httptest mock that echoes success regardless of the filter.

src/cli/client.go
```diff
@@ -304,6 +306,9 @@ func (c *Client) doRawRequest(method string, segments ...string) ([]byte, error)
 		escaped = append(escaped, url.PathEscape(segment))
```
jymaire, line 306: url.Values.Encode() percent-encodes the brackets in filters[namespace][EQUALS] to filters%5Bnamespace%5D%5BEQUALS%5D; if the server's filter binder matches on the raw (non-decoded) parameter name, the filter is dropped and the search falls back to all-namespace results.

endpoint becomes .../secrets?filters%5Bnamespace%5D%5BEQUALS%5D=my.namespace. A binder that inspects the raw query string rather than decoded param names would not recognize the key, returning the unfiltered secret list. This is the concrete mechanism behind the filter finding and is not covered by the unit test (which only asserts kestractl SENDS the param, via r.URL.Query() which decodes it, not that the server applies it).

src/cli/secrets.go
fdelbrayelle: Verified against a live Kestra EE instance (see the QA comment on this PR) — the server does honor `filters[namespace][EQUALS]`. Confirmed namespace isolation directly: created secrets in `system` and a separate `qa.namespace`, and `secrets list system` stayed empty while `secrets list qa.namespace` showed only its own secret. Not a theoretical risk, exercised end-to-end.

src/cli/client.go
fdelbrayelle: Same live-EE QA covers this: the raw request logged was `.../secrets?filters%5Bnamespace%5D%5BEQUALS%5D=system` (i.e. exactly the percent-encoded form `url.Values.Encode()` produces), and the server applied it correctly — isolation confirmed between `system` and `qa.namespace`. So the binder does decode the param name; no drop-to-unfiltered behavior observed.

jymaire (approved):

PR #178, opened 2026-09-18
fix: executions run --wait taskRunList + secrets set --from-file

### LINKED ISSUE #173: executions run --wait returns an empty taskRunList
`executions run --wait --output json` returns the execution with an empty `taskRunList`, whatever actually ran. You have to fetch the execution again to see the task runs.

Related: `executions list` is not newest-first and does not show sub-executions.

Together these made a working `Loop` look like it had done nothing. Loop iterations run as sub-executions, so the parent legitimately holds only the loop task and the child logs live elsewhere. With `--wait` returning an empty `taskRunList` and `list` not surfacing sub-executions, there was no obvious path from "my loop ran" to "here is what it did". I spent a couple of hours convinced I had found a bug in `Loop` before proving otherwise.

Would help: `--wait` returning the task runs, and `list` defaulting to newest-first with sub-executions visible.

Importance: medium. `--wait` is what scripts use, and its output currently makes a correct execution look empty. Urgency: low.

Environment:
- kestractl 2.2.0
- Kestra 2.0.0 (commit 7c22ca8, 2026-09-07), Enterprise
- macOS, arm64

### LINKED ISSUE #174: secrets set: a value starting with "-" is read as a flag, and $(cat) breaks SSH keys silently
Two things hit while loading an SSH private key into a secret.

**1.** `kestractl secrets set MY_KEY "-----BEGIN OPENSSH PRIVATE KEY-----..."` fails until you put `--` before the value. Private keys and PEM certificates all start with `-`, so this is easy to hit.

**2.** `"$(cat id_ed25519)"` strips the trailing newline OpenSSH requires. The secret stores and resolves fine, and the failure only appears later at use time as `error in libcrypto: unsupported`. Nothing in that points at a missing newline.

A `--from-file` flag that stores the file byte for byte would avoid both.

Importance: low, but the second one is expensive to diagnose because the error appears far from the cause. Urgency: low.

Environment:
- kestractl 2.2.0
- Kestra 2.0.0 (commit 7c22ca8, 2026-09-07), Enterprise
- macOS, arm64

fdelbrayelle:
## Summary
- `executions run --wait` re-fetches the execution once it completes and surfaces `taskRunList`, so callers no longer need a second `executions get` to see what ran. Fixes #173.
- `secrets set` gains `--from-file` to store a file's bytes verbatim, avoiding leading-`-` flag parsing and shell `$(cat)` trailing-newline stripping. Fixes #174.

## Test plan
- [x] `go build ./...`
- [x] `go test ./src/...`
- [ ] QA against a live Kestra EE instance (in progress)


fdelbrayelle:
## QA report

Ran against a live Kestra EE dev instance (`kestra-core-run`, Docker, `europe-west1-docker.pkg.dev/.../kestra-ee:develop`, server reported `2.1.0-SNAPSHOT`, commit `8176490`).

### #173 — `executions run --wait` taskRunList

Created a 1-task flow (`qa.kestractl:kestractl-qa-173`) and ran it with `--wait`.

- **Before the fix**: `--wait --output json` returned the execution with no `taskRunList` key at all.
- **First attempt at the fix** (plain re-fetch via `Execution()`) still came back empty against this server build — the SDK's strict unmarshal of the GET response errors on an unrelated missing/mismatched required field, so the typed path silently produced nothing. Caught this via `--verbose`, which showed the raw GET response *did* contain a populated `taskRunList` even though the SDK call returned an error.
- **Fixed** by falling back to `tryParseExecutionFromError` (same pattern already used elsewhere in this file for this class of SDK/API mismatch) to pull `taskRunList` from the raw JSON when the typed call errors.
- **Verified**, both output modes:
  - `--output json`: `taskRunList` present with `id`/`taskId`/`state` for the task run.
  - table output: new "Task Runs:" section listing `- hello (<id>): SUCCESS`.

### #174 — `secrets set` dash-prefixed values / `--from-file`

- `secrets set qa.kestractl KEY "-----BEGIN..."` (no `--`) still fails with `bad flag syntax` — expected, this is standard POSIX flag parsing and the documented workaround (`--` or `--from-file`) still applies.
- `secrets set qa.kestractl KEY -- "-----BEGIN..."` — succeeds.
- `secrets set qa.kestractl KEY --from-file <path>` — succeeds; unit test (`TestSecretsSetCommand_FromFilePreservesTrailingNewline`) confirms the exact file bytes (including trailing `\n`) are sent as the secret value.
- `secrets set qa.kestractl KEY value --from-file <path>` (both given) → `--from-file cannot be combined with a positional value`.
- `secrets set qa.kestractl KEY` (neither given) → `a value is required unless --from-file is set`.

All QA secrets/flow created during this pass were deleted afterward.

### Side finding (not in scope for this PR, filing separately)

`kestractl config add/use/remove` ignore the `--config` flag entirely and always read/write `~/.kestractl/config.yaml`, regardless of `--config`. Hit this while trying to use an isolated config for QA — first attempt clobbered a real `~/.kestractl/config.yaml` that had existed since July. Worked around it for this QA pass with an isolated `$HOME`. Will file as its own issue.

### Build/test

- `go build ./...` clean
- `go test ./src/...` passing (includes new tests for both fixes)

fdelbrayelle:
## QA against live Kestra EE (Docker, `develop`, via `kestra-core-run --ee --docker`)

All three fixes verified end-to-end against a running instance (`localhost:8080`).

### #173 — `executions run --wait` returns `taskRunList`
Deployed a 2-task flow, ran `executions run <ns> <flow> --wait -o json`. Output includes both task runs with state, no second `executions get` needed:
```json
"taskRunList": [
  {"id": "<taskrun-id-1>", "state": "SUCCESS", "taskId": "first"},
  {"id": "<taskrun-id-2>", "state": "SUCCESS", "taskId": "second"}
]
```

### #174 — `secrets set --from-file`
Wrote a 41-byte file whose first line looks like a PEM header (starts with `-`) and which ends with a trailing newline. `secrets set qa.kestractl179 QA_FROM_FILE --from-file <file>` succeeded with **no `--` needed** despite the leading dash. Ran a flow reading the secret through `wc -c`: log output was exactly `41` — byte-for-byte match, trailing newline preserved (the `$(cat)` gotcha this flag exists for).

### #179 — `config add/use/remove/show` ignored `--config`
Added a dedicated commit fixing this on the same branch.
Repro with `HOME` pointed at an empty fake home and `--config /tmp/some-other-config.yaml`:
- Before: context silently written to `$HOME/.kestractl/config.yaml` instead.
- After: `config add`/`use`/`remove` all write to the `--config` path; `$HOME/.kestractl` is never created. Verified the resulting config file actually authenticates against the live EE instance (`namespaces list` returned real data), then exercised `use` and `remove` against the same isolated file.

Build (`go build ./...`) and unit tests (`go test ./src/...`) pass.

src/cli/executions.go
```diff
@@ -1026,6 +1026,29 @@ func runExecutionsRun(client *Client, namespace, flowID string, wait bool, input
...
+		if id, ok := execution["id"].(string); ok && id != "" {
+			full, _, ferr := client.API.ExecutionsAPI.Execution(client.Ctx, id, client.Tenant).Execute()
+			switch {
+			case ferr == nil && full != nil:
+				execution["taskRunList"] = taskRunListToMaps(full.GetTaskRunList())
+			case ferr != nil:
+				// Same SDK/API type-mismatch class as tryParseExecutionFromError's
+				// other callers: the raw success body still has taskRunList even
+				// when the SDK's strict unmarshal errors on an unrelated field.
+				if raw := tryParseExecutionFromError(ferr); raw != nil {
```
jymaire: A genuine re-fetch failure is silently swallowed here. This `case ferr != nil` only recovers a `taskRunList` when `tryParseExecutionFromError` finds an id-bearing body (the SDK type-mismatch class). For a real failure — network error, 403, 5xx — `tryParseExecutionFromError` returns `nil`, so the block does nothing: no `taskRunList`, no diagnostic. The user then sees `State: SUCCESS` with no Task Runs section and can't tell whether the flow had zero tasks or whether the follow-up GET failed.

Suggest emitting a one-line warning to stderr in that fall-through so the missing data is at least explained.

src/cli/executions.go
```diff
@@ -1037,10 +1060,62 @@ func runExecutionsRun(client *Client, namespace, flowID string, wait bool, input
...
 		}
+		printTaskRunList(w, execution)
 		return nil
 	})
 }
 
+// taskRunListToMaps converts SDK TaskRun entries into the map shape used for
+// both JSON output and table rendering.
+func taskRunListToMaps(taskRuns []kestra.TaskRun) []map[string]any {
+	list := make([]map[string]any, len(taskRuns))
```
jymaire, line 1081: `make([]map[string]any, len(taskRuns))` returns a non-nil empty slice for zero task runs, so `--wait` always emits `"taskRunList": []` in JSON while a non-`--wait` run omits the key entirely — field presence becomes flag-dependent, which trips consumers that key on presence to decide whether task runs were fetched.

Returning `nil` when the list is empty (and skipping the `execution["taskRunList"] = ...` assignment in that case) would make the field omit cleanly and consistently.

jymaire (approved):

fdelbrayelle:
Both addressed in 6f45dbd: re-fetch failures after `--wait` now warn on stderr instead of failing silently, and `taskRunList` is omitted (not `[]`) when there are no task runs, regardless of `--wait`.

